PR #20, opened 2026-03-27
Show thinking indicator inline in assistant response bubble

yagudaev:
## Summary
- **Fixes NAN-573**: The thinking/loading indicator now renders inside an assistant response bubble in the FlatList instead of as a separate element below the list
- Animated typing dots (3 pulsing circles) replace the old `ActivityIndicator + "Thinking..."` text
- When streaming starts, the thinking bubble transitions to the streamed text in the same position — no layout shift or scroll jump

## What changed
- Added a `THINKING_MESSAGE_ID` (-2) sentinel to inject a placeholder assistant message into `displayMessages` when `isThinking` is true
- Created a `ThinkingDots` component using `Animated` opacity pulses on 3 dots with staggered delays
- Updated `MessageBubble` to detect the thinking placeholder and render `ThinkingDots` instead of text content
- Removed the standalone `isThinking` block that was rendered outside the FlatList

## Test plan
- [ ] Start a voice call, speak, and verify the thinking dots appear inline as an assistant bubble (not a separate element)
- [ ] Verify no layout shift when assistant starts responding — the bubble stays in place
- [ ] Send a text message and verify the thinking dots appear inline before streaming begins
- [ ] Verify the dots animate smoothly (3 pulsing circles with staggered timing)
- [ ] Verify `npx tsc --noEmit` passes cleanly

### DIFF
--- a/app/(tabs)/index.tsx
+++ b/app/(tabs)/index.tsx
@@ -20,10 +20,11 @@ import { Stack } from 'expo-router'
 import { MicIcon, MicOffIcon, PhoneOffIcon, PlusIcon, RefreshCwIcon, SendIcon, XIcon } from 'lucide-react-native'
 import { useColorScheme } from 'nativewind'
 import { useCallback, useEffect, useRef, useState } from 'react'
-import { ActivityIndicator, FlatList, Image, KeyboardAvoidingView, Platform, Pressable, View } from 'react-native'
+import { ActivityIndicator, Animated, FlatList, Image, KeyboardAvoidingView, Platform, Pressable, View } from 'react-native'
 
 const MD_IMAGE_REGEX = /!\[([^\]]*)\]\(([^)]+)\)/g
 const URL_IMAGE_REGEX = /(?:^|\s)(https?:\/\/\S+\.(?:png|jpg|jpeg|gif|webp)(?:\?\S*)?)/gi
+const THINKING_MESSAGE_ID = -2
 
 const VOICE_SYSTEM_PROMPT = `\
 You are on a live voice call. Keep responses concise and conversational — short \
@@ -689,9 +690,15 @@ export default function ChatScreen() {
     setIsMuted(newMuted)
   }, [isMuted])
 
-  const displayMessages = streamingText !== null
-    ? [...messages, { id: -1, conversation_id: conversationId ?? 0, role: 'assistant' as const, content: streamingText, created_at: Date.now(), stt_latency_ms: null, llm_latency_ms: null, tts_latency_ms: null }]
-    : messages
+  const displayMessages = (() => {
+    let msgs = messages as Message[]
+    if (streamingText !== null) {
+      msgs = [...msgs, { id: -1, conversation_id: conversationId ?? 0, role: 'assistant' as const, content: streamingText, created_at: Date.now(), stt_latency_ms: null, llm_latency_ms: null, tts_latency_ms: null }]
+    } else if (isThinking) {
+      msgs = [...msgs, { id: THINKING_MESSAGE_ID, conversation_id: conversationId ?? 0, role: 'assistant' as const, content: '', created_at: Date.now(), stt_latency_ms: null, llm_latency_ms: null, tts_latency_ms: null }]
+    }
+    return msgs
+  })()
 
   const { partials } = transcriptBuffer
 
@@ -742,15 +749,6 @@ export default function ChatScreen() {
           ) : null
         }
       />
-
-      {isThinking && (
-        <View className="flex-row items-center gap-2 px-4 py-2">
-          <View className="flex-row items-center gap-2 rounded-2xl rounded-bl-sm bg-muted px-4 py-3">
-            <ActivityIndicator size="small" color="#888" />
-            <Text className="text-sm text-muted-foreground">Thinking...</Text>
-          </View>
-        </View>
-      )}
 
       {reconnectState.status === 'reconnecting' && (
         <View className="items-center gap-2 border-t border-border bg-muted/50 px-4 py-3">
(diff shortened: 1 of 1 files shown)


app/(tabs)/index.tsx
```diff
@@ -689,9 +690,15 @@ export default function ChatScreen() {
...
   }, [isMuted])
 
-  const displayMessages = streamingText !== null
-    ? [...messages, { id: -1, conversation_id: conversationId ?? 0, role: 'assistant' as const, content: streamingText, created_at: Date.now(), stt_latency_ms: null, llm_latency_ms: null, tts_latency_ms: null }]
-    : messages
+  const displayMessages = (() => {
+    let msgs = messages as Message[]
+    if (streamingText !== null) {
+      msgs = [...msgs, { id: -1, conversation_id: conversationId ?? 0, role: 'assistant' as const, content: streamingText, created_at: Date.now(), stt_latency_ms: null, llm_latency_ms: null, tts_latency_ms: null }]
+    } else if (isThinking) {
```
yagudaev: **Animation loop sync issue**: using `Animated.delay` inside `Animated.loop` means the delay is re-applied on every iteration, not just the first. The three loops run independently so the dots will stay correctly staggered each cycle (0ms, 200ms, 400ms head-start per loop), but the total per-loop duration differs per dot: dot1 = 800ms, dot2 = 1000ms, dot3 = 1200ms. Over time they drift out of phase.

A more reliable approach is to use a single shared timing value with interpolated output ranges, or use a single loop that sequences all three dots explicitly. For example:

```ts
// Single loop, explicit sequence
Animated.loop(
  Animated.sequence([
    Animated.timing(dot1, { toValue: 1, duration: 300, useNativeDriver: true }),
    Animated.timing(dot1, { toValue: 0.3, duration: 300, useNativeDriver: true }),
    Animated.timing(dot2, { toValue: 1, duration: 300, useNativeDriver: true }),
    // ...
  ])
)
```

Or use a single `Animated.Value` with `interpolate` for each dot offset. This guarantees they stay in sync forever.

app/(tabs)/index.tsx
```diff
@@ -689,9 +690,15 @@ export default function ChatScreen() {
     setIsMuted(newMuted)
   }, [isMuted])
 
-  const displayMessages = streamingText !== null
-    ? [...messages, { id: -1, conversation_id: conversationId ?? 0, role: 'assistant' as const, content: streamingText, created_at: Date.now(), stt_latency_ms: null, llm_latency_ms: null, tts_latency_ms: null }]
-    : messages
+  const displayMessages = (() => {
```
yagudaev: Minor style nit: the IIFE (`(() => { ... })()`) adds ceremony without benefit here. A simple early-return pattern or useMemo reads more naturally:

```ts
let displayMessages: Message[] = messages
if (streamingText !== null) {
  displayMessages = [...messages, { id: -1, ... }]
} else if (isThinking) {
  displayMessages = [...messages, { id: THINKING_MESSAGE_ID, ... }]
}
```

This avoids the IIFE and keeps the pattern consistent with the rest of the file.

app/(tabs)/index.tsx
```diff
@@ -838,6 +836,53 @@ function PartialBubble({ role, text }: { role: 'user' | 'assistant', text: strin
   )
 }
 
+function ThinkingDots() {
+  const dot1 = useRef(new Animated.Value(0.3)).current
+  const dot2 = useRef(new Animated.Value(0.3)).current
+  const dot3 = useRef(new Animated.Value(0.3)).current
+
+  useEffect(() => {
```
yagudaev, line 841: Minor: `LatencyBadge` is rendered for the thinking placeholder message (id=-2) in the `renderItem`. It returns `null` immediately since all latency fields are null, so nothing is visible — but the component mounts/unmounts on every thinking state change.

Consider guarding against it in `renderItem` or in `LatencyBadge` itself by checking for the sentinel id. Alternatively, since `ThinkingDots` renders a completely separate early-return in `MessageBubble`, the `LatencyBadge` could simply be skipped when `item.id === THINKING_MESSAGE_ID`:

```tsx
renderItem={({ item }) => (
  <>
    <MessageBubble message={item} />
    {item.id !== THINKING_MESSAGE_ID && <LatencyBadge message={item} />}
  </>
)}
```

app/(tabs)/index.tsx
```diff
@@ -838,6 +836,53 @@ function PartialBubble({ role, text }: { role: 'user' | 'assistant', text: strin
...
+    a1.start()
+    a2.start()
+    a3.start()
+
+    return () => {
+      a1.stop()
+      a2.stop()
+      a3.stop()
+    }
+  }, [dot1, dot2, dot3])
```
yagudaev, line 858: The `backgroundColor: '#888'` is hardcoded and will look wrong in light mode (too dark) vs. dark mode (might be fine). The old thinking indicator used `bg-muted` + `text-muted-foreground` via nativewind for theme-awareness.

Consider using `useColorScheme` + a conditional or pulling a color from your theme tokens instead of a hardcoded hex.

yagudaev (commented):
The core approach is solid — injecting a sentinel message into `displayMessages` is the right fix for NAN-573 and cleanly eliminates the layout shift. The `ThinkingDots` animation is a nice UX upgrade over the old `ActivityIndicator + "Thinking..."` pattern.

One issue worth fixing before merge:

**Animation correctness (needs a fix)**:
The three `Animated.loop` instances run independently with different total cycle times (dot1=800ms, dot2=1000ms, dot3=1200ms due to the `Animated.delay` being part of each loop). The dots will drift out of phase after several cycles. Use a single loop with an explicit sequence, or a single `Animated.Value` with `interpolate` for each dot's offset, to keep them permanently synchronized.

**Minor (non-blocking but worth addressing)**:
- The IIFE for `displayMessages` is unnecessary — a flat `let` + `if/else if` block is simpler and consistent with the rest of the file.
- `LatencyBadge` is still mounted in `renderItem` for the thinking placeholder (returns `null`, so no visual issue, but it's a no-op mount/unmount).
- `backgroundColor: '#888'` in `ThinkingDots` is hardcoded and doesn't follow the existing `bg-muted` theming — will look different in light vs. dark mode compared to the bubble it sits inside.

PR #23, opened 2026-03-27
Fix Custom Pipeline: echo cancellation, assistant messages, TTS errors

yagudaev:
## Summary

- **Echo fix**: Switch audio session to `.voiceChat` mode for hardware echo cancellation; add `pendingTTSCount` + `isStreamComplete` tracking so STT only restarts after ALL TTS chunks finish (not between each sentence)
- **Assistant messages**: Add `onAssistantResponse` event that emits the full LLM response so JS can save it to conversation history
- **Partial transcript side**: Fix partial transcripts showing on the wrong (assistant/left) side of chat — they're user speech so they should be on the right
- **TTS errors in UI**: Add `onError` callback to `TTSProvider.speak()` protocol; ElevenLabs (and OpenAI) now surface HTTP errors, missing API keys, etc. directly as messages in the chat UI
- **Thinking state**: Clear thinking indicator on `onAssistantResponse` so it doesn't hang forever if TTS fails

## Test plan

- [ ] Start a Custom Pipeline call with Apple TTS — verify no echo (your own voice not transcribed back)
- [ ] Have a multi-sentence exchange — verify STT only restarts after the full response is spoken
- [ ] Verify assistant responses appear in conversation history after the call
- [ ] Speak something — verify partial transcript appears on the right (user) side
- [ ] Configure ElevenLabs with an invalid API key or free-tier library voice — verify error message appears in chat UI
- [ ] Configure ElevenLabs with a valid owned voice — verify TTS speaks correctly

### DIFF
--- a/app/(tabs)/index.tsx
+++ b/app/(tabs)/index.tsx
@@ -85,6 +85,7 @@ export default function ChatScreen() {
   const [isThinking, setIsThinking] = useState(false)
   const [isConnecting, setIsConnecting] = useState(false)
   const [streamingText, setStreamingText] = useState<string | null>(null)
+  const [streamingRole, setStreamingRole] = useState<'user' | 'assistant'>('assistant')
   const flatListRef = useRef<FlatList<Message>>(null)
   const hasScrolledRef = useRef(false)
   const { playJoin, playEnd, startThinking, stopThinking } = useCallSounds()
@@ -430,6 +431,7 @@ export default function ChatScreen() {
     streamCompletion(allMessages, apiKey, model, apiUrl, systemPrompt, conversationId, {
       onToken: (text) => {
         setIsThinking(false)
+        setStreamingRole('assistant')
         setStreamingText(text)
       },
       onDone: async (text) => {
@@ -625,6 +627,7 @@ export default function ChatScreen() {
     if (ttsProvider === 'elevenlabs') {
       const elKey = await getSetting('elevenlabs_api_key')
       const elVoice = await getSetting('elevenlabs_voice_id')
+      console.log('[CustomPipeline] ElevenLabs config — key:', elKey ? `${elKey.substring(0, 8)}...` : 'MISSING', 'voice:', elVoice || 'default')
       if (elKey) ttsConfig.apiKey = elKey
       if (elVoice) ttsConfig.voiceId = elVoice
     } else if (ttsProvider === 'openai') {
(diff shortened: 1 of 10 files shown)


yagudaev:
## Review Summary

Adds assistant response emission and TTS error surfacing end‑to‑end. JS shows live user partials, writes assistant final on event, and wires new native events. Core pipeline introduces `pendingTTSCount` + `isStreamComplete` gating and per‑stage latency. Overall good direction — a few issues to address before merge.

---

### 🔴 Major Issues

**1. ElevenLabs audio session likely breaks echo‑cancellation**

`configureAudioSessionForPlayback` (ElevenLabsTTSProvider.swift:154) uses mode `.default`, overriding the `.voiceChat` mode that `AppleSTTProvider` set for hardware AEC. This re‑enables echo on every ElevenLabs speak call.

```swift
// Current (breaks AEC):
try audioSession.setCategory(.playAndRecord, mode: .default, ...)
// Fix: keep voiceChat so hardware AEC stays active
try audioSession.setCategory(.playAndRecord, mode: .voiceChat, options: [.defaultToSpeaker, .allowBluetooth])
```

**2. OpenAI TTS error surfacing incomplete**

Network errors, non‑200 HTTP responses, empty data, and AVAudio decode errors all call `onComplete()` but never `onError()`. The UI won't show anything when OpenAI TTS fails. ElevenLabs already does this correctly — OpenAI should match.

- OpenAITTSProvider.swift:68 — request error path: no `onError` call
- OpenAITTSProvider.swift:77 — empty data path: no `onError` call  
- OpenAITTSProvider.swift:116 — AVAudio decode error: no `onError` call
- Missing HTTP status check entirely (ElevenLabs has one at line 87)

**3. ElevenLabs TTS: concurrent speaks overwrite `audioPlayer`**

When the LLM streams multiple sentences quickly, `speakText()` is called for each before the first finishes playing. ElevenLabs starts a new network request per sentence and overwrites `audioPlayer` when the response arrives. The previous player loses its delegate — its `onComplete` never fires — so `pendingTTSCount` leaks upward and STT never restarts.

Apple TTS (`AVSpeechSynthesizer`) queues utterances natively and doesn't have this problem.

---

### 🟡 Concurrency & Races

**`isStreamComplete` set on TCP close, not on `[DONE]`**

`handleSSEData` (PipelineManager.swift:153) flushes the buffer when it sees `[DONE]`, but `isStreamComplete` is set in `handleSSEComplete` (PipelineManager.swift:182–183), which fires on URLSession task completion. If the TCP connection closes quickly this is fine, but it's fragile. Setting `isStreamComplete = true` when `[DONE]` is seen makes the restart logic deterministic.

**`pendingTTSCount` can go negative**

If a provider ever calls `onComplete` after `stop()` clears the counter, `pendingTTSCount` goes negative. A simple clamp is defensive:

```swift
// PipelineManager.swift:235
pendingTTSCount = max(0, pendingTTSCount - 1)
```

---

### 🟡 Error Handling

- ElevenLabs: AVAudio decode error (line 140) calls `onCompleteCallback` but not `onErrorCallback` — inconsistent with the rest of the error paths.
- Log level nit: `configureAudioSessionForPlayback` uses `logger.error` for a success message ("configured for playback") — should be `logger.debug`.

---

### ✅ What Looks Good

- `streamingRole` fix correctly shows partial user transcripts on the right side.
- `onAssistantResponse` event and DB save wiring is clean.
- `onError` → `sendEvent("onError")` propagation end-to-end works.
- No semicolons in TS/JS. Helpers at bottom. TypeScript types updated.
- `restartSTTIfReady` guard is correct — won't restart mid-TTS.

---

**Decision: Request changes** — primarily the ElevenLabs audio session mode regression and OpenAI error surfacing gaps.

PR #24, opened 2026-03-27
Fix masked input fields: tap-to-focus and dot alignment (NAN-587, NAN-589)

yagudaev:
## Summary
- **NAN-587:** Masked API key inputs were not focusable on tap because `editable={false}` was set when hidden. Replaced manual bullet-character rendering with native `secureTextEntry`, so the TextInput is always editable and focusable regardless of mask state.
- **NAN-589:** Masked dots were misaligned with the eye icon. Added `leading-[40px]` to the TextInput for vertical centering, and gave the eye-icon Pressable explicit height + `items-center justify-center` so both elements share the same vertical center.

## Changes
- `SecretInput` now uses `secureTextEntry={!visible}` instead of rendering `\u2022` characters manually
- TextInput is always editable and receives `onChangeText` in both states
- Eye icon Pressable has explicit `h-10` height with centering and `hitSlop` for better touch target

## Test plan
- [ ] Open Settings, scroll to any masked API key field (e.g. Vapi API Key, ElevenLabs API Key, OpenClaw API Key)
- [ ] Tap directly on the masked input field -- keyboard should appear and field should focus
- [ ] Type a value -- it should appear as masked dots
- [ ] Tap the eye icon -- value should be revealed
- [ ] Tap the eye icon again -- value should be masked again
- [ ] Verify the masked dots and eye icon are vertically centered/aligned
- [ ] Test on iOS device or Simulator

### DIFF
--- a/app/(tabs)/settings.tsx
+++ b/app/(tabs)/settings.tsx
@@ -335,18 +335,22 @@ function SecretInput({
   return (
     <View className="flex-row items-center rounded-md border border-input bg-background dark:bg-input/30">
       <TextInput
-        className="h-10 min-w-0 flex-1 px-3 text-base text-foreground"
+        className="h-10 min-w-0 flex-1 px-3 text-base leading-[40px] text-foreground"
         placeholder={placeholder}
         placeholderTextColor="#888"
-        value={visible ? value : value ? '\u2022'.repeat(Math.min(value.length, 30)) : ''}
-        onChangeText={visible ? onChangeText : undefined}
-        editable={visible}
+        value={value}
+        onChangeText={onChangeText}
+        secureTextEntry={!visible}
         autoCapitalize="none"
         autoCorrect={false}
         numberOfLines={1}
         scrollEnabled
       />
-      <Pressable onPress={() => setVisible((prev) => !prev)} className="shrink-0 px-3">
+      <Pressable
+        onPress={() => setVisible((prev) => !prev)}
+        className="h-10 shrink-0 items-center justify-center px-3"
+        hitSlop={{ top: 8, bottom: 8, left: 4, right: 8 }}
+      >
         <Icon
           as={visible ? EyeOffIcon : EyeIcon}
           size={20}
--- a/app/(tabs)/settings.tsx
+++ b/app/(tabs)/settings.tsx
@@ -335,7 +335,7 @@ function SecretInput({
   return (
     <View className="flex-row items-center rounded-md border border-input bg-background dark:bg-input/30">
       <TextInput
-        className="h-10 min-w-0 flex-1 px-3 text-base leading-[40px] text-foreground"
+        className="h-10 min-w-0 flex-1 px-3 py-2 text-base text-foreground"
         placeholder={placeholder}
         placeholderTextColor="#888"
         value={value}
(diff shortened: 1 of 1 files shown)


app/(tabs)/settings.tsx
```diff
@@ -335,18 +335,22 @@ function SecretInput({
   return (
     <View className="flex-row items-center rounded-md border border-input bg-background dark:bg-input/30">
       <TextInput
-        className="h-10 min-w-0 flex-1 px-3 text-base text-foreground"
+        className="h-10 min-w-0 flex-1 px-3 text-base leading-[40px] text-foreground"
```
yagudaev: The `leading-[40px]` class maps to `lineHeight: 40` in React Native. On iOS, when `secureTextEntry={true}` is active, the OS renders its own bullet/dot characters and does **not** respect the `lineHeight` style — it can actually cause the dots to appear vertically clipped or shifted in certain iOS versions (known RN issue: facebook/react-native#18534 and related).

Consider removing `leading-[40px]` from the TextInput class and instead relying on the row container's existing `items-center` to handle vertical alignment. The `h-10` on both the TextInput and the Pressable should be sufficient to keep them aligned without needing an explicit line-height override.

app/(tabs)/settings.tsx
```diff
@@ -335,18 +335,22 @@ function SecretInput({
...
         autoCapitalize="none"
         autoCorrect={false}
         numberOfLines={1}
         scrollEnabled
       />
-      <Pressable onPress={() => setVisible((prev) => !prev)} className="shrink-0 px-3">
+      <Pressable
+        onPress={() => setVisible((prev) => !prev)}
+        className="h-10 shrink-0 items-center justify-center px-3"
+        hitSlop={{ top: 8, bottom: 8, left: 4, right: 8 }}
```
yagudaev: Minor nit: the `hitSlop` left value (`4`) is asymmetric with the right (`8`). Since the Pressable already has `px-3` (12px each side), the hitSlop is more of a safety net than a functional need — but if keeping it, matching left/right (e.g., both 8) would feel more intentional. Not a blocker.

yagudaev (commented):
Overall this is a clean, correct fix for both tickets. The core approach -- switching to native secureTextEntry instead of the manual bullet-character hack -- is exactly right and elegantly solves NAN-587 (focus bug) while also being the idiomatic iOS solution.

One actionable concern: leading-[40px] (i.e. lineHeight: 40) on a TextInput with secureTextEntry is a known footgun on iOS -- the OS-rendered password dots ignore lineHeight and can appear clipped or misaligned. Worth verifying on device before merging, and likely safer to remove it and rely on h-10 + row container items-center for alignment instead (see inline comment).

Everything else looks good:
- Eye icon toggle: no regression -- secureTextEntry={!visible} correctly flips on the same state variable
- Code style: no semicolons, SecretInput helper stays at the bottom, no formatting sprawl
- Empty field: fine -- value='' with secureTextEntry shows no dots, placeholder renders normally
- Long API keys: native secureTextEntry handles arbitrary length correctly (no more Math.min(..., 30) cap, which was silently truncating display for keys longer than 30 chars)
- Paste: native secureTextEntry supports paste natively on iOS -- actual improvement over the old approach

PR #25, opened 2026-03-27
Add latency display toggle in chat and detailed stats in settings (NAN-584)

yagudaev:
## Summary
- Add a stopwatch icon toggle in the chat header that shows/hides per-message latency badges (STT, LLM, TTS). The toggle persists across sessions via the `show_latency` setting and is off by default.
- Enhance the Settings "Latency Stats" section to show min/max alongside averages for each pipeline stage (STT, LLM, TTS, Total), with turn count.
- Add refresh (reload stats) and clear (null out latency data with confirmation) buttons to the latency stats card.
- Add `clearLatencyData()` DB function that nulls latency columns without deleting messages.

## Test plan
- [ ] Open the Chat tab and verify the stopwatch icon appears in the header (dimmed/inactive)
- [ ] Tap the stopwatch icon -- it should turn amber and latency badges should appear under messages that have latency data
- [ ] Kill and reopen the app -- the toggle state should persist
- [ ] Open Settings and scroll to "Latency Stats" -- verify avg, min, and max are shown for each stage
- [ ] Tap the refresh icon to reload stats
- [ ] Tap the trash icon and confirm the alert clears latency data
- [ ] Run `npx tsc --noEmit` -- no type errors

### DIFF
--- a/app/(tabs)/index.tsx
+++ b/app/(tabs)/index.tsx
@@ -401,7 +401,7 @@ export default function ChatScreen() {
   }, [])
 
   useEffect(() => {
-    getSetting('show_latency').then((v) => { if (v === 'true') setShowLatency(true) })
+    getSetting('show_latency').then((v) => { if (v === 'true') setShowLatency(true) }).catch(() => {})
   }, [])
 
   useEffect(() => { loadMessages() }, [loadMessages])
--- a/app/(tabs)/index.tsx
+++ b/app/(tabs)/index.tsx
@@ -3,7 +3,7 @@ import { Button } from '@/components/ui/button'
 import { Icon } from '@/components/ui/icon'
 import { Input } from '@/components/ui/input'
 import { Text } from '@/components/ui/text'
-import { addMessage, createConversation, getConversation, getLatestConversation, getMessages, getSetting, updateConversationVapi, type Message, type LatencyData } from '@/db'
+import { addMessage, createConversation, getConversation, getLatestConversation, getMessages, getSetting, setSetting, updateConversationVapi, type Message, type LatencyData } from '@/db'
 import { getApiConfig, streamCompletion } from '@/lib/chat'
 import { compactMessages } from '@/lib/compact'
 import { useConversationContext } from '@/lib/conversation-context'
@@ -17,7 +17,7 @@ import type { FinalTranscriptEvent, LatencyUpdateEvent } from '@/modules/expo-cu
 import ExpoVapiModule from '@/modules/expo-vapi'
 import type { FunctionCallEvent, SpeechEvent, TranscriptEvent } from '@/modules/expo-vapi'
 import { Stack } from 'expo-router'
-import { MicIcon, MicOffIcon, PhoneOffIcon, PlusIcon, RefreshCwIcon, SendIcon, XIcon } from 'lucide-react-native'
+import { MicIcon, MicOffIcon, PhoneOffIcon, PlusIcon, RefreshCwIcon, SendIcon, TimerIcon, XIcon } from 'lucide-react-native'
 import { useColorScheme } from 'nativewind'
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { ActivityIndicator, Animated, FlatList, Image, KeyboardAvoidingView, Platform, Pressable, View } from 'react-native'
@@ -85,6 +85,7 @@ export default function ChatScreen() {
   const [isThinking, setIsThinking] = useState(false)
   const [isConnecting, setIsConnecting] = useState(false)
   const [streamingText, setStreamingText] = useState<string | null>(null)
+  const [showLatency, setShowLatency] = useState(false)
   const flatListRef = useRef<FlatList<Message>>(null)
   const hasScrolledRef = useRef(false)
   const { playJoin, playEnd, startThinking, stopThinking } = useCallSounds()
@@ -399,6 +400,10 @@ export default function ChatScreen() {
     })()
   }, [])
 
+  useEffect(() => {
+    getSetting('show_latency').then((v) => { if (v === 'true') setShowLatency(true) })
+  }, [])
+
   useEffect(() => { loadMessages() }, [loadMessages])
 
   const sendMessage = useCallback(async () => {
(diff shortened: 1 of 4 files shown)


db/messages.ts
```diff
@@ -68,13 +76,29 @@ export async function getLatencyAverages(): Promise<LatencyAverages> {
...
     turn_count: number
   }>(
     `SELECT
       AVG(stt_latency_ms) as avg_stt,
       AVG(llm_latency_ms) as avg_llm,
       AVG(tts_latency_ms) as avg_tts,
       AVG(COALESCE(stt_latency_ms, 0) + COALESCE(llm_latency_ms, 0) + COALESCE(tts_latency_ms, 0)) as avg_total,
+      MIN(stt_latency_ms) as min_stt,
+      MAX(stt_latency_ms) as max_stt,
+      MIN(llm_latency_ms) as min_llm,
```
yagudaev, line 96: **Bug: `min_total`/`max_total` include 0 for missing stages, skewing the result.**

A row that only has `llm_latency_ms` set (e.g. a Vapi turn where STT/TTS are null) contributes `0 + llm + 0` to the MIN/MAX computation, dragging min down toward 0. The same issue exists in `avg_total` from the original code, but it is especially misleading here since a single partially-instrumented turn can dominate min/max.

Individual stage `MIN`/`MAX` (e.g. `MIN(stt_latency_ms)`) work correctly because SQLite ignores NULLs automatically. The `COALESCE` trick for total breaks that.

Consider filtering to rows where all three stages are present, or use a CASE expression:
```sql
MIN(CASE WHEN stt_latency_ms IS NOT NULL AND llm_latency_ms IS NOT NULL AND tts_latency_ms IS NOT NULL
     THEN stt_latency_ms + llm_latency_ms + tts_latency_ms END) as min_total,
MAX(CASE WHEN stt_latency_ms IS NOT NULL AND llm_latency_ms IS NOT NULL AND tts_latency_ms IS NOT NULL
     THEN stt_latency_ms + llm_latency_ms + tts_latency_ms END) as max_total
```

app/(tabs)/settings.tsx
```diff
@@ -291,22 +291,53 @@ export default function SettingsScreen() {
...
+                  }}
+                  className="p-2"
+                >
+                  <Icon as={Trash2Icon} size={18} className="text-destructive" />
+                </Pressable>
+              )}
+            </View>
+          </View>
           {loadingStats ? (
             <ActivityIndicator size="small" color="#888" />
```
yagudaev, line 327: **Missing `await` on `loadLatencyStats()` after clear — and no error handling for the reload.**

```tsx
onPress: async () => {
  await clearLatencyData()
  loadLatencyStats()  // not awaited
},
```

The reload is fire-and-forget, which is probably fine UX-wise (it will run), but if `loadLatencyStats` throws, the error is silently swallowed. Also, the trash icon stays visible until `loadLatencyStats` finishes and `latencyStats.turnCount` drops to 0 — this is fine, just a note that the button is briefly tappable again during the reload.

Minor suggestion: either `await loadLatencyStats()` so errors propagate correctly, or wrap in a `try/catch`. Low priority.

app/(tabs)/index.tsx
```diff
@@ -707,9 +718,14 @@ export default function ChatScreen() {
       <Stack.Screen
```
yagudaev, line 741: **`headerRight` is a new inline arrow function on every render — `showLatency` changes now cause more frequent re-creation.**

```tsx
headerRight: () => (   // new function reference every render
  <View ...>
    <Pressable onPress={toggleLatencyDisplay} ...>
```

`toggleLatencyDisplay` is already memoized with `useCallback`, but `headerRight` itself is a fresh lambda every render, so `Stack.Screen` options object changes on every render. This was already true for the `PlusIcon` button, but now that `showLatency` (and the dynamic color `#f59e0b` vs `#666`) is captured, every toggle immediately recreates it.

This is unlikely to cause visible jank (header re-renders are cheap), but if you want to be precise, `useMemo` the options object or extract to a separate component:
```tsx
const headerRight = useCallback(() => (
  <View className="mr-2 flex-row items-center">
    ...
  </View>
), [showLatency, colorScheme, toggleLatencyDisplay, startNewConversation])
```
Not blocking, just flagging since the FlatList render performance was already a concern.

app/(tabs)/index.tsx
```diff
@@ -399,6 +400,10 @@ export default function ChatScreen() {
     })()
   }, [])
 
+  useEffect(() => {
```
yagudaev, line 404: **`getSetting` rejection is unhandled — if the DB read fails, the promise rejects silently.**

```tsx
useEffect(() => {
  getSetting('show_latency').then((v) => { if (v === 'true') setShowLatency(true) })
}, [])
```

No `.catch()` attached. A DB read failure on startup will produce an unhandled promise rejection. At minimum add `.catch(console.warn)`. The rest of the codebase uses `async/await` with try/catch for settings — this pattern is inconsistent.

```tsx
useEffect(() => {
  getSetting('show_latency')
    .then((v) => { if (v === 'true') setShowLatency(true) })
    .catch((err) => console.warn('[Chat] Failed to load show_latency setting:', err))
}, [])
```

app/(tabs)/settings.tsx
```diff
@@ -432,13 +463,26 @@ function OptionGroup<T extends string>({
```
yagudaev, line 467: **Duplicate `// --- Helper Components ---` section marker.**

The file now has two `// --- Helper Components ---` markers — one at the original location (line ~432 in the base, before `SegmentedControl`/`OptionGroup`) and this new one added before `LatencyStageSection`. The old `LatencyStatRow` was placed after the second marker, and the new `LatencyStageSection` is added in the same spot, so the duplication is carried forward.

Either remove this second marker, or consolidate all helper components under the single `// --- Helper Components ---` section. The `formatLatencyMs` helper at the bottom should move under `// --- Helper Functions ---` which is correct, but having two component sections is confusing.

yagudaev (commented):
Overall the PR is clean and well-scoped — the toggle, persistence, and clear function all work correctly. One real bug to fix before merging, plus a few minor issues.

**Must fix:**
- **SQL min_total/max_total bug** (see inline comment on db/messages.ts): COALESCE(x, 0) turns NULLs into 0 before MIN/MAX, so a row that only has one stage populated (e.g. a Vapi turn with only LLM latency) drags the minimum down incorrectly. Individual stage min/max are fine because SQLite MIN/MAX ignore NULLs automatically. Fix with a CASE WHEN expression.

**Should fix:**
- Unhandled promise rejection in the show_latency load useEffect — no .catch() on the getSetting call.
- loadLatencyStats() not awaited after clearLatencyData() in the Alert handler.

**Nits (will not block):**
- Duplicate `// --- Helper Components ---` section marker in settings.tsx.
- headerRight is a bare inline arrow in Stack.Screen options so it re-creates on every render — already true before this PR but more impactful now. useCallback on it would be clean.

PR #26, opened 2026-03-27
Autosave settings on change, remove save button (NAN-588)

yagudaev:
## Summary
- Settings now auto-save as the user changes them -- no explicit save button needed
- Toggles, dropdowns, and segmented controls persist immediately via `saveImmediate`
- Text inputs are debounced at 500ms via `saveDebounced` to avoid saving on every keystroke
- A subtle animated "Saved" checkmark indicator replaces the old save button and alert dialog
- A `loadedRef` guard prevents spurious saves during initial hydration from the database
- New reusable `useAutoSave` hook in `lib/use-auto-save.ts`

## Test plan
- [ ] Open Settings tab, toggle Voice Mode between Vapi and Custom -- verify it persists after closing and reopening the tab
- [ ] Change a text field (e.g., API URL) and wait ~500ms -- verify "Saved" indicator appears briefly
- [ ] Type rapidly in a text field -- verify only one save occurs after typing stops
- [ ] Kill and relaunch the app -- verify all settings are preserved
- [ ] Verify no save occurs on initial screen load (no spurious "Saved" flash)

### DIFF
--- a/app/(tabs)/settings.tsx
+++ b/app/(tabs)/settings.tsx
@@ -3,7 +3,7 @@ import { Icon } from '@/components/ui/icon'
 import { Input } from '@/components/ui/input'
 import { Text } from '@/components/ui/text'
 import { getSetting, getLatencyAverages, type LatencyAverages } from '@/db'
-import { useAutoSave } from '@/lib/use-auto-save'
+import { useAutoSave, type SaveStatus } from '@/lib/use-auto-save'
 import { CheckIcon } from 'lucide-react-native'
 import { EyeIcon, EyeOffIcon } from 'lucide-react-native'
 import { useCallback, useEffect, useRef, useState } from 'react'
@@ -484,7 +484,7 @@ function OptionGroup<T extends string>({
   )
 }
 
-function SavedIndicator({ status }: { status: import('@/lib/use-auto-save').SaveStatus }) {
+function SavedIndicator({ status }: { status: SaveStatus }) {
   const opacity = useRef(new Animated.Value(0)).current
 
   useEffect(() => {
--- a/app/(tabs)/settings.tsx
+++ b/app/(tabs)/settings.tsx
@@ -1,12 +1,13 @@
-import { Button } from '@/components/ui/button'
 import { Card } from '@/components/ui/card'
 import { Icon } from '@/components/ui/icon'
 import { Input } from '@/components/ui/input'
 import { Text } from '@/components/ui/text'
-import { getSetting, setSetting, getLatencyAverages, type LatencyAverages } from '@/db'
+import { getSetting, getLatencyAverages, type LatencyAverages } from '@/db'
+import { useAutoSave } from '@/lib/use-auto-save'
+import { CheckIcon } from 'lucide-react-native'
 import { EyeIcon, EyeOffIcon } from 'lucide-react-native'
-import { useCallback, useEffect, useState } from 'react'
-import { ActivityIndicator, Alert, KeyboardAvoidingView, Platform, Pressable, ScrollView, TextInput, View } from 'react-native'
+import { useCallback, useEffect, useRef, useState } from 'react'
+import { ActivityIndicator, Animated, KeyboardAvoidingView, Platform, Pressable, ScrollView, TextInput, View } from 'react-native'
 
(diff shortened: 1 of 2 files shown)


lib/use-auto-save.ts
```diff
@@ -0,0 +1,77 @@
...
+    timerRefs.current.set(key, setTimeout(() => {
+      timerRefs.current.delete(key)
+      persistSetting(key, value)
+    }, debounceMs))
+  }, [debounceMs, persistSetting])
+
+  return { saveStatus, saveImmediate, saveDebounced }
+}
+
+export type { SaveStatus }
```
yagudaev, line 77: **Bug: status goes idle too early, cutting off the fade-out animation.**

The hook resets `saveStatus` to `'idle'` after 1500ms. But `SavedIndicator` needs the status to stay `'saved'` for the full animation duration: 200ms fade-in + 1000ms hold + 300ms fade-out = **1500ms**. When `idle` is set at exactly 1500ms, the component returns `null` and unmounts before the 300ms fade-out completes.

Simplest fix — bump the hook timer to 1800ms (1500ms animation + 300ms buffer):
```ts
fadeTimerRef.current = setTimeout(() => {
  setSaveStatus('idle')
  fadeTimerRef.current = null
}, 1800)  // was 1500
```

app/(tabs)/settings.tsx
```diff
@@ -83,31 +87,83 @@ export default function SettingsScreen() {
...
 
-  const handleSave = async () => {
-    await setSetting('vapi_api_key', vapiApiKey)
-    await setSetting('assistant_id', assistantId)
-    await setSetting('default_model', defaultModel)
-    await setSetting('openclaw_api_key', openclawApiKey)
-    await setSetting('openclaw_api_url', openclawApiUrl)
-    await setSetting('system_prompt', systemPrompt)
-
-    // Save voice pipeline settings
```
yagudaev, line 97: **Early-return conflicts with the fade-out animation.**

```tsx
if (status === 'idle') return null
```

When `saveStatus` flips to `'idle'`, the component unmounts instantly — the Animated.timing fade-out never completes. The `opacity.setValue(0)` branch in the useEffect is dead code because the unmount fires first.

Two clean options:
1. Remove the early return and always render the `Animated.View` — let opacity handle visibility (starts at 0, invisible until needed). This is the simplest fix.
2. Keep the early return but add a `'hiding'` intermediate state in the hook so the component stays mounted until the fade-out animation finishes before the hook transitions to `'idle'`.

app/(tabs)/settings.tsx
```diff
@@ -83,31 +87,83 @@ export default function SettingsScreen() {
       if (oaiKey) setOpenaiTtsApiKey(oaiKey)
       const oaiVoice = await getSetting('openai_tts_voice')
       if (oaiVoice) setOpenaiTtsVoice(oaiVoice)
```
yagudaev, line 88: Minor nit: the inline import type `import('@/lib/use-auto-save').SaveStatus` is unnecessarily verbose. `SaveStatus` is already exported from the module — just add it to the named imports:
```tsx
import { useAutoSave, type SaveStatus } from '@/lib/use-auto-save'
```

yagudaev (commented):
Overall this is a clean, well-structured implementation. The hook abstraction is solid, the loadedRef guard is correct, cleanup on unmount is handled, and the Save button is properly removed. I found one functional bug with the SavedIndicator animation being cut off mid-fade, and a couple of minor issues noted inline.

app/(tabs)/settings.tsx
```diff
@@ -83,31 +87,83 @@ export default function SettingsScreen() {
...
+
+  const updateElevenlabsVoiceId = useCallback((v: string) => {
+    setElevenlabsVoiceId(v)
+    if (loadedRef.current) saveDebounced('elevenlabs_voice_id', v)
+  }, [saveDebounced])
+
+  const updateOpenaiTtsApiKey = useCallback((v: string) => {
+    setOpenaiTtsApiKey(v)
+    if (loadedRef.current) saveDebounced('openai_tts_api_key', v)
+  }, [saveDebounced])
```
yagudaev, line 165: Seems ugly, is there no better way to do this?

PR #28, opened 2026-03-27
Implement Kokoro on-device TTS provider (NAN-586)

yagudaev:
## Summary

- **KokoroTTSProvider**: Full implementation using `mlalma/kokoro-ios` MLX Swift library
  - Downloads model (`kokoro-v1_0.safetensors`) + voices (`voices.npz`) from HuggingFace on first use
  - Caches files in app's caches directory — no re-download after first run
  - Synthesizes audio via `KokoroTTS.generateAudio()` on a background thread
  - Plays back via `AVAudioEngine` + `AVAudioPlayerNode` (mono, 24kHz)
  - Guarded with `#if canImport(KokoroSwift)` — graceful fallback on iOS < 18 or if package not linked
- **ExpoCustomPipelineModule**: extracted `makeSTTProvider`/`makeTTSProvider` helpers to bottom; added `"kokoro"` case with `@available(iOS 18.0, *)` guard
- **Settings UI**: added "Kokoro On-Device (iOS 18+)" as a TTS provider option
- **Xcode project**: `KokoroSwift` SPM package added to `VoiceClaw` target

## Notes
- First use triggers a model download (~300-600MB) — no UI progress indicator yet
- Requires iOS 18+ (iPhone 12 Pro running iOS 18 is the test device ✓)
- Voice used: `af_heart` (American female) from the bundled voices.npz

## Test plan
- [ ] Go to Settings → TTS Provider → select "Kokoro On-Device (iOS 18+)"
- [ ] Start a custom pipeline call — first use will download the model (may take a minute on WiFi)
- [ ] Subsequent calls should use cached model and speak immediately
- [ ] On a device running iOS < 18, selecting Kokoro should surface an error via the pipeline's onError event

### DIFF
--- a/app/(tabs)/settings.tsx
+++ b/app/(tabs)/settings.tsx
@@ -4,6 +4,7 @@ import { Input } from '@/components/ui/input'
 import { Text } from '@/components/ui/text'
 import { getSetting, getLatencyAverages, clearLatencyData, type LatencyAverages } from '@/db'
 import { useAutoSave, type SaveStatus } from '@/lib/use-auto-save'
+import ExpoCustomPipelineModule from '@/modules/expo-custom-pipeline/src/ExpoCustomPipelineModule'
 import { CheckIcon, EyeIcon, EyeOffIcon, RefreshCwIcon, Trash2Icon } from 'lucide-react-native'
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { ActivityIndicator, Alert, Animated, KeyboardAvoidingView, Platform, Pressable, ScrollView, TextInput, View } from 'react-native'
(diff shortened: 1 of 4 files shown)


yagudaev:
PR Review — focused on KokoroTTSProvider.swift, ExpoCustomPipelineModule.swift, and settings.tsx.

**KokoroTTSProvider.swift**
- AVAudioEngine: Lazy engine/player creation and main-mixer connection look good. Using \`.interrupts\` on \`scheduleBuffer\` is reasonable for cutting over mid-utterance. Consider \`eng.prepare()\` before \`eng.start()\` to reduce start latency.
- Download/cache: Solid — caches to \`cachesDirectory\`, sequential downloads, temp-to-final move, cancel support.
- **Potential bug:** \`defaultVoiceKey\` is \`"af_heart.npy"\` (line 94). Many NPZ readers expose keys without the \`.npy\` extension (e.g., \`"af_heart"\`). If \`voices.npz\` uses bare keys, the lookup will silently fail. Suggest trying both with a fallback.
- Minor nit: Reusing a single \`onCompleteCallback\` across overlapping speaks can misattribute completion if an interrupted buffer fires its handler. Likely OK with \`.interrupts\`, but worth confirming.

**ExpoCustomPipelineModule.swift**
- Helper extraction keeps the module clean. iOS 18 gating for Kokoro is correct.
- Suggest emitting an \`onError\` event (not just printing) when an unknown/unavailable provider is set so JS can react.

**app/(tabs)/settings.tsx**
- \`openaiTtsVoice\` state is typed as \`string\` but should be \`typeof OPENAI_TTS_VOICES[number]\`. Also validate against the list when loading from storage.

**Verdict:** Approve with two fixes:
1. Guard the Kokoro voice key lookup against bare vs \`.npy\` key format
2. Tighten OpenAI voice type to the literal union

Great work overall!

PR #32, opened 2026-03-28
Add API key validation on Settings page (NAN-593)

yagudaev:
## Summary
- Add a "Test" button next to each cloud API key input (OpenClaw, ElevenLabs, Deepgram, OpenAI TTS, Vapi) on the Settings page
- Each button makes a lightweight API call to verify the key is valid, showing a spinner while testing, green checkmark if valid, or red error icon with message if invalid
- Validation status resets automatically when the key value changes (cache until key changes)
- New `lib/validate-api-key.ts` module with provider-specific validation functions

## Test plan
- [ ] Open Settings, enter a valid OpenClaw API key, tap Test -- should show green checkmark
- [ ] Enter an invalid API key for any provider, tap Test -- should show red error icon and error message
- [ ] Change the API key after testing -- status should reset to "Test" button
- [ ] Test with empty key -- Test button should be disabled (dimmed)
- [ ] Verify all 5 providers: OpenClaw, ElevenLabs, Deepgram, OpenAI TTS, Vapi
- [ ] Verify the app builds successfully in iOS simulator: `npx expo run:ios --device "iPhone 16 Pro"`

### DIFF
--- a/app/(tabs)/settings.tsx
+++ b/app/(tabs)/settings.tsx
@@ -378,9 +378,6 @@ export default function SettingsScreen() {
                 value={vapiPublicKey}
                 onChangeText={updateVapiPublicKey}
                 placeholder="Enter your Vapi public key"
-                validationStatus={validationStatus.vapi}
-                validationError={validationErrors.vapi}
-                onTest={() => testApiKey('vapi', vapiPublicKey)}
               />
             </View>
 
--- a/app/(tabs)/settings.tsx
+++ b/app/(tabs)/settings.tsx
@@ -6,8 +6,9 @@ import { Text } from '@/components/ui/text'
 import { getSetting, getLatencyAverages, clearLatencyData, type LatencyAverages } from '@/db'
 import { runPipelineTests, type TestResult } from '@/lib/pipeline-test-runner'
 import { useAutoSave, type SaveStatus } from '@/lib/use-auto-save'
+import { validateApiKey, type Provider, type ValidationStatus } from '@/lib/validate-api-key'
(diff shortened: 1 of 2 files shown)


yagudaev (commented):
**Code Review — NAN-593 API Key Validation**

Good overall shape — architecture is clean, state management is correct, no keys are logged anywhere, the `idle → testing → valid/invalid` flow reads naturally, and code style (no semicolons, helpers at bottom of file) is correct. One real bug needs fixing before merging.

---

## Bug: Vapi public key vs private key

`validateVapi` calls `https://api.vapi.ai/assistant` with `Authorization: Bearer <key>`. That REST endpoint requires a **private API key**, but the field being validated here is `vapi_public_key` — the client-side token used to initialise the Vapi SDK. Sending the public key to the REST API will return 401 even for a perfectly valid key, so this test will always report invalid for users with correct credentials.

**Recommended fix:** Don't pass `onTest` to the Vapi `SecretInput`. The public key is only validated implicitly by the SDK at call-start time; there is no lightweight REST call that accepts it. The `onTest` prop is already optional so this is a one-line change:

```diff
-              validationStatus={validationStatus.vapi}
-              validationError={validationErrors.vapi}
-              onTest={() => testApiKey('vapi', vapiPublicKey)}
```

---

## Minor notes (non-blocking)

**`validateOpenClaw` — 400 treated as valid**
The comment explains the heuristic. Fine for now, but worth a `// TODO` noting that a 400 from OpenRouter is ambiguous (could be model-not-found with valid key, or a malformed request).

**Race condition on rapid re-test**
If the user taps Test, changes the key, and taps Test again before the first fetch resolves, the first response can overwrite the second. An `AbortController` ref (one per provider) would fix it cleanly. Low priority for a settings screen but worth a follow-up ticket.

PR #40, opened 2026-03-28
Refactor voice pipeline to hybrid TS + Swift AudioCoordinator (NAN-599)

yagudaev:
## Summary
- New `AudioCoordinator` in Swift owns STT, TTS, and VAD lifecycle
- Barge-in is atomic on the native side: detect voice → stop TTS → start STT with no JS bridge round trip
- `ExpoCustomPipelineModule` delegates to AudioCoordinator instead of calling providers directly
- Replaces `startBargeInDetection/stopBargeInDetection` with single `setBargeInEnabled(bool)` 
- TS side simplified: removed `bargeInActiveRef`, `onBargeIn` handler just cancels LLM stream

## Test plan
- [ ] Single turn: speak → assistant responds → TTS plays → STT restarts
- [ ] Multi-turn: sequential turns work
- [ ] Barge-in: speak during TTS → TTS stops, STT restarts atomically
- [ ] Tap-to-interrupt: mic button stops TTS and starts listening

### DIFF
--- a/lib/pipeline-test-runner.ts
+++ b/lib/pipeline-test-runner.ts
@@ -61,7 +61,7 @@ export async function runPipelineTests(
   // Cleanup
   ExpoCustomPipelineModule.stopListening()
   ExpoCustomPipelineModule.stopSpeaking()
-  ExpoCustomPipelineModule.stopBargeInDetection()
+  ExpoCustomPipelineModule.setBargeInEnabled(false)
 
   return results
 }
--- a/lib/use-pipeline.ts
+++ b/lib/use-pipeline.ts
@@ -42,20 +42,19 @@ export function usePipeline(callbacks: PipelineCallbacks): PipelineControls {
   const sentenceBufferRef = useRef('')
   const stopCompletionRef = useRef<(() => void) | null>(null)
   const isActiveRef = useRef(false)
-  const bargeInActiveRef = useRef(false)
   const activeTurnIdRef = useRef(0)
 
-  // ── Latency tracking ──────────────────────────────────────────────
+  // -- Latency tracking --------------------------------------------------
   //
-  // STT latency:  time from startListening() call → onFinalTranscript event
-  //               (mic opens → user utterance fully recognised)
+  // STT latency:  time from startListening() call -> onFinalTranscript event
+  //               (mic opens -> user utterance fully recognised)
   //
-  // LLM latency:  time from API request sent → first streamed token received
+  // LLM latency:  time from API request sent -> first streamed token received
   //               (network + model time-to-first-token)
   //
-  // TTS latency:  time from speak() call → onTTSStart event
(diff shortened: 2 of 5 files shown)


yagudaev (commented):
Good overall direction -- the architecture is clean and the atomicity goal is right. Found one real bug and a few minor issues.

## Bug: VAD never starts for the first TTS sentence

In `AudioCoordinator.speak()`:

```swift
onStart: { [weak self] in
    guard let self = self else { return }
    self.delegate?.audioCoordinatorDidStartTTS()  // fires onTTSStart to JS
    if self.bargeInEnabled {                       // still false here
        self.startBargeInMonitoring()
    }
}
```

`delegate?.audioCoordinatorDidStartTTS()` sends `onTTSStart` to JS asynchronously over the bridge. By the time JS receives it and calls `setBargeInEnabled(true)` back across the bridge, the `if self.bargeInEnabled` check has already executed and returned `false`. VAD never starts for the first sentence.

For subsequent sentences `bargeInEnabled` is already `true` (set by the first `onTTSStart`), so VAD works fine for them.

**Fix**: Have `setBargeInEnabled(true)` itself start the detector if TTS is already playing:

```swift
func setBargeInEnabled(_ enabled: Bool) {
    bargeInEnabled = enabled
    if enabled && ttsProvider?.isSpeaking == true {
        startBargeInMonitoring()
    } else if !enabled {
        bargeInDetector.stopMonitoring()
    }
}
```

And remove the `if self.bargeInEnabled { self.startBargeInMonitoring() }` from the `onStart` closure entirely.

## Minor: `onBargeIn` handler bypasses `isActiveRef` guard

After barge-in, native already called `coordinator.startListening()`. If the user calls `stop()` (sets `isActiveRef = false`) and a latent `onFinalTranscript` fires from that native STT session, `callAPI()` will execute regardless. The old code called `startListening()` from JS which was blocked by the `isActiveRef` guard. The surface area for this pre-existing race increased slightly.

## Minor: `configureAudioSession()` is called three times per barge-in

`AudioCoordinator.startListening()`, `BargeInDetector.startMonitoring()`, and `AppleSTTProvider.beginRecognition()` each configure the audio session independently. Not a correctness issue, just redundant log spam.

## Minor: `sttProviderName` / `ttsProviderName` in `ExpoCustomPipelineModule` are now dead code

They're set but never read after delegating to `AudioCoordinator`. Safe to remove.

## What's good

- `weak var delegate` correctly prevents retain cycles
- `handleBargeIn()` is atomic: stopTTS -> stopVAD -> startSTT -> emit, all on main thread (BargeInDetector dispatches to main before firing)
- `setBargeInEnabled` is a cleaner API than the start/stop pair
- `bargeInActiveRef` removal is correct -- guard is now native-side
- `onBargeIn` no longer calls `startListening()` from JS -- right call, avoids double-start
- No semicolons in TS, helpers at bottom -- clean

PR #41, opened 2026-03-28
Add MVP OpenClaw connection mode selector (HTTP vs Plugin/WebSocket)

yagudaev:
## Summary
- **Settings UI**: Added a segmented control in the OpenClaw Configuration card to toggle between HTTP (existing REST/SSE) and Plugin (WebSocket) connection modes. HTTP mode shows the existing URL/API key/model fields. Plugin mode shows gateway URL, auth token, live connection status bar with connect/disconnect button, and setup instructions.
- **WebSocket LLM client** (`lib/ws-completion.ts`): New module that manages WebSocket lifecycle (connect/disconnect/status), auto-reconnect with exponential backoff (up to 5 attempts), and streaming completion via `req:agent` / `event:agent` protocol format. Exposes the same `onToken/onDone/onError` callback interface as `streamCompletion`.
- **Unified LLM abstraction** (`lib/chat.ts`): Added `unifiedStreamCompletion` that checks the `openclaw_connection_mode` setting and delegates to either HTTP or WebSocket path. `getApiConfig` now returns `connectionMode`. Existing `streamCompletion` callers are unaffected — migration to unified function can happen incrementally.
- **Plugin scaffold** (`openclaw-plugin/`): Directory with `package.json`, `openclaw.plugin.json` manifest, `index.ts` entry point, and `src/channel.ts` channel implementation with documented protocol and a stub response for end-to-end testing.

## Notes
- The WebSocket protocol format (`req:agent` / `event:agent`) is a best guess based on the channel naming convention. Comments in the code document what may need adjustment once tested against a real OpenClaw gateway.
- The `openclaw` CLI was not found on the system, so protocol details could not be verified.
- Existing HTTP callers (`streamCompletion`) are not changed — they continue to work as before. The new `unifiedStreamCompletion` is opt-in for callers that want automatic mode switching.

## Test plan
- [ ] Open Settings, verify OpenClaw config card shows "HTTP" and "Plugin (WebSocket)" segmented control
- [ ] Toggle to HTTP mode: verify URL, API key, and model fields appear
- [ ] Toggle to Plugin mode: verify gateway URL, auth token, connection status bar, and setup instructions appear
- [ ] In Plugin mode, tap Connect without a gateway URL: verify error state shown
- [ ] Toggle between modes and verify settings persist after closing and reopening the app
- [ ] Verify existing HTTP chat flow still works unchanged
- [ ] TypeScript compiles cleanly (`npx tsc --noEmit`)

### DIFF
--- a/app/(tabs)/settings.tsx
+++ b/app/(tabs)/settings.tsx
@@ -560,7 +560,7 @@ export default function SettingsScreen() {
 
               <WsConnectionStatusBar
                 status={wsStatus}
-                onConnect={connectWs}
+                onConnect={() => connectWs(gatewayUrl, authToken)}
                 onDisconnect={disconnectWs}
               />
 
--- a/app/(tabs)/settings.tsx
+++ b/app/(tabs)/settings.tsx
@@ -4,11 +4,13 @@ import { Icon } from '@/components/ui/icon'
 import { Input } from '@/components/ui/input'
 import { Text } from '@/components/ui/text'
 import { getSetting, setSetting, getLatencyAverages, type LatencyAverages } from '@/db'
+import type { OpenClawConnectionMode } from '@/lib/chat'
 import { runPipelineTests, type TestResult } from '@/lib/pipeline-test-runner'
 import { useAutoSave, type SaveStatus } from '@/lib/use-auto-save'
 import { validateApiKey, type Provider, type ValidationStatus } from '@/lib/validate-api-key'
+import { connectWs, disconnectWs, getWsStatus, addWsStatusListener, type WsConnectionStatus } from '@/lib/ws-completion'
 import ExpoCustomPipelineModule from '@/modules/expo-custom-pipeline/src/ExpoCustomPipelineModule'
-import { AlertCircleIcon, CheckIcon, EyeIcon, EyeOffIcon, PlayIcon, RefreshCwIcon } from 'lucide-react-native'
+import { AlertCircleIcon, CheckIcon, EyeIcon, EyeOffIcon, PlayIcon, RefreshCwIcon, WifiIcon, WifiOffIcon } from 'lucide-react-native'
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { ActivityIndicator, Animated, KeyboardAvoidingView, Platform, Pressable, ScrollView, Switch, TextInput, View } from 'react-native'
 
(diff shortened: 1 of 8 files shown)


yagudaev (commented):
Good foundational work — the module structure is clean, code style is consistent (no semicolons, helpers at the bottom), no tokens are logged, and the provisional protocol is well-documented with NOTE comments. A few issues need addressing before merge:

## Blocking issues

### 1. Plugin mode is never actually used in the chat screen (`app/(tabs)/index.tsx`)

`index.tsx` still imports and calls `streamCompletion` directly. `unifiedStreamCompletion` was added to `lib/chat.ts` but is never wired up. A user who selects Plugin mode in Settings will keep hitting the HTTP path — the feature is silently broken for real chat usage.

The fix is straightforward: swap the `streamCompletion` call in `index.tsx` for `unifiedStreamCompletion` and pass `connectionMode` from `getApiConfig()` (which now returns it). Same for `lib/use-pipeline.ts` and `lib/pipeline-test-runner.ts` if they should also respect the mode.

**File:** `app/(tabs)/index.tsx` line 582
```ts
// Before
const { apiKey, model, apiUrl } = await getApiConfig()
streamCompletion(allMessages, apiKey, model, apiUrl, systemPrompt, conversationId, { ... })

// After
const { apiKey, model, apiUrl, connectionMode } = await getApiConfig()
unifiedStreamCompletion(allMessages, apiKey, model, apiUrl, systemPrompt, conversationId, connectionMode, { ... })
```

### 2. Race condition: Connect button reads stale settings from DB

`connectWs` (called directly from the Connect button) reads `openclaw_gateway_url` and `openclaw_auth_token` from the DB via `getSetting`. The inputs use a 500 ms debounced save. If a user types a URL and immediately taps Connect, the DB may still hold the old value and the connection will fail silently with just an `error` status and no message to the user.

Two options: either flush the debounced save before connecting, or pass the current field values into `connectWs` as parameters rather than having it re-read from the DB.

## Minor / non-blocking issues

### 3. Incomplete cancellation of in-progress WS completion

`wsStreamCompletion` returns a cancel function that nulls `activeCallbacks`, but does **not** send a cancel/abort message to the gateway. This means the gateway will keep streaming tokens that are silently dropped. For a scaffold this is acceptable, but it should be documented with a `// TODO: send cancel message to gateway` comment so it isn't forgotten.

### 4. `onclose` reconnect logic edge case

In `openSocket` → `ws.onclose`, after `disconnectWs()` sets `reconnectAttempts = MAX_RECONNECT_ATTEMPTS` and calls `ws.close()`, the `onclose` fires asynchronously. If `disconnectWs` races with an in-flight reconnect that hasn't incremented `reconnectAttempts` yet, the `else if (reconnectAttempts >= MAX_RECONNECT_ATTEMPTS)` branch could fire and flip status back to `'error'` even after a manual disconnect. Minor, worth noting.

### 5. `getWsConfig()` is exported but appears unused

`getWsConfig` in `ws-completion.ts` (line 123) is exported but nothing imports it. Dead code — either use it or remove it.

### 6. Model field hidden in Plugin mode with no way to configure

When `connectionMode === 'plugin'`, the Model field is hidden. `wsStreamCompletion` still receives `model` from the caller, and since the field is invisible the fallback `'openclaw:voice'` always applies. Fine for MVP, but worth a comment.

## Summary

Structurally solid: module separation is clean, UI conditional rendering is correct, auto-reconnect with exponential backoff is well-implemented, and security is fine (no token logging). The primary blocker is that plugin mode is never invoked from actual chat — `unifiedStreamCompletion` needs to replace the `streamCompletion` call in `index.tsx`.

PR #55, opened 2026-04-14
Fix OpenAI response race condition (NAN-618)

yagudaev:
## Summary
- add response lifecycle tracking and queued response creation in the OpenAI realtime adapter
- cancel in-flight responses before resuming after tool results to avoid active-response collisions
- add a focused regression test covering queued creates, cancel/resume behavior, and watchdog deferral

## Test plan
- [x] `npx tsx relay-server/test/test-openai-response-guard.ts`
- [x] `npx tsc --noEmit -p relay-server/tsconfig.json`

Refs NAN-618

### DIFF
--- a/relay-server/src/adapters/openai.ts
+++ b/relay-server/src/adapters/openai.ts
@@ -419,6 +419,7 @@ export class OpenAIAdapter implements ProviderAdapter {
       // Errors
       case "error":
         logError(`[openai] Error: ${event.error?.message ?? event}`)
+        this.resetResponseState()
         this.sendToClient?.({
           type: "error",
           message: event.error?.message ?? "upstream error",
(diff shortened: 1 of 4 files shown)


yagudaev (commented):
Claude Code review completed.

Addressed findings in follow-up commits:
- deduplicated pending `response.cancel` sends while a cancel is already in flight
- moved test helpers to the bottom of the file and removed ASI-style leading semicolon casts
- closed the remaining send/ack race by marking the response active immediately after a successful `response.create` send
- reset the response state machine on upstream `error` events so future turns can recover
- added regression coverage for cancel dedupe and error-state recovery

Verification after fixes:
- `npx tsx relay-server/test/test-openai-response-guard.ts`
- `npx tsc --noEmit -p relay-server/tsconfig.json`

PR #57, opened 2026-04-14
Adapt OpenClaw prompts for OpenAI realtime (NAN-620)

yagudaev:
## Summary
- add an OpenAI-only identity transform that compresses OpenClaw soul/identity content into voice-first prompt sections
- keep the Gemini path on the existing identity prompt so only OpenAI realtime gets the tailored rewrite
- add a regression test that validates the OpenAI transform and guards the Gemini fallback behavior

## Test plan
- [x] `npx tsx relay-server/test/test-openai-instructions-transform.ts`
- [x] `npx tsc --noEmit -p relay-server/tsconfig.json`

Refs NAN-620

### DIFF
--- a/relay-server/src/instructions.ts
+++ b/relay-server/src/instructions.ts
@@ -203,9 +203,18 @@ function compactLines(lines: Array<string | null>, limit: number): string[] {
 function extractSection(markdown: string | null, title: string): string {
   if (!markdown) return ""
 
-  const escapedTitle = escapeRegex(title)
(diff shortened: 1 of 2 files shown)


yagudaev (commented):
Claude Code review completed.

Addressed findings in follow-up commits:
- replaced the invalid JavaScript `\Z` end-of-string regex in `extractSection()` with a line-based section parser
- removed duplicated safety guidance so high-risk lines only appear in the safety section

Latest Claude pass: no blocking issues; one non-blocking note about tolerating leading whitespace on markdown headings.

Verification after fixes:
- `npx tsx relay-server/test/test-openai-instructions-transform.ts`
- `npx tsc --noEmit -p relay-server/tsconfig.json`

PR #66, opened 2026-04-16
Add 1007 to Gemini reconnectable close codes (NAN-628)

yagudaev:
## Summary
- Add 1007 to RECONNECTABLE_CLOSE_CODES so the relay retries instead of dying
- Drop stale video frames on reconnect instead of flushing them (likely cause of 1007)

Ref: NAN-628

## Dependency
None — independent of PR #65

## Test plan
- [ ] Start voice call with screen sharing
- [ ] Wait for goAway rotation (~10 min)
- [ ] Verify session reconnects cleanly without 1007

### DIFF
--- a/relay-server/src/adapters/gemini.ts
+++ b/relay-server/src/adapters/gemini.ts
@@ -161,7 +161,7 @@ export class GeminiAdapter implements ProviderAdapter {
 
   /**
    * Handle an unexpected upstream close after the session was established.
-   * For reconnectable codes (1001/1006/1011/1012/1013) with a resumption handle
+   * For reconnectable codes (1001/1006/1007/1012/1013) with a resumption handle
    * available, silently re-open the session. Otherwise surface an error.
    */
   private handleUpstreamClose(code: number) {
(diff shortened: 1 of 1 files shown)


relay-server/src/adapters/gemini.ts
```diff
@@ -16,9 +16,11 @@ const RECONNECT_BACKOFF_MS = 500
...
 // Gemini's graceful end-of-session path is goAway → we reconnect proactively
 // before the close lands. These codes cover transport-level drops (network,
-// server restart, rate limit). 1011 is intentionally excluded: it usually
-// signals a server error that's not helped by replaying the resumption handle.
-const RECONNECTABLE_CLOSE_CODES = new Set([1001, 1006, 1012, 1013])
+// server restart, rate limit). 1007 is included because Gemini can send it after
+// a goAway rotation, likely triggered by stale video frames flushed into the new
+// session (invalid argument). 1011 is intentionally excluded: it usually signals
+// a server error that's not helped by replaying the resumption handle.
+const RECONNECTABLE_CLOSE_CODES = new Set([1001, 1006, 1007, 1012, 1013])
```
yagudaev, line 23: Can you use const names instead of magic numbers?

PR #70, opened 2026-04-16
Reduce HistoryPage DB polling

yagudaev:
## Summary
- Skip DB fetch when document/tab is hidden
- Increase polling interval from 2s to 10s

## Test plan
- [ ] Switch to History tab — conversations load
- [ ] Start a call, send messages, switch to History — new conversation appears within 10s
- [ ] Switch to Chat tab — History stops polling (check via console/network)

### DIFF
--- a/desktop/src/renderer/src/App.tsx
+++ b/desktop/src/renderer/src/App.tsx
@@ -49,7 +49,7 @@ export function App() {
             <ChatPage />
           </div>
           <div className={`flex-1 flex flex-col overflow-hidden ${activeTab !== 'history' ? 'hidden' : ''}`}>
-            <HistoryPage />
+            <HistoryPage isVisible={activeTab === 'history'} />
           </div>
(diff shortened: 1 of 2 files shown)


yagudaev:
**Bug found + fixed** (committed to this branch):

`document.hidden` only fires when the Electron **window** is minimized/hidden — not when the user switches between in-app tabs. Since `App.tsx` keeps all pages mounted and hides them with a CSS `hidden` class, `document.hidden` stays `false` the entire time the window is open. The polling was still firing every 10s on all tabs.

**Fix:**
- Added `isVisible: boolean` prop to `HistoryPage`, driven by `activeTab === 'history'` in `App.tsx`
- Polling interval is now only started when `isVisible` is true (and cleared immediately when the tab is hidden)
- Added an immediate `loadConversations()` call when `isVisible` transitions to `true`, so the list is always fresh when you switch to History — no need to wait up to 10s

**Other items reviewed:**
- 10s interval: reasonable for this use case
- `clearInterval` cleanup on unmount: was already correct; the new per-`isVisible` effect also cleans up properly when visibility changes

PR #154, opened 2026-04-18
fix(desktop): rebuild better-sqlite3 against electron ABI on install

yagudaev:
## Problem

After PR #149 bumped \`better-sqlite3\` to 12.9.0, the desktop app fails on launch with:

\`\`\`
The module 'better_sqlite3.node' was compiled against a different Node.js version
using NODE_MODULE_VERSION 141. This version of Node.js requires NODE_MODULE_VERSION 145.
\`\`\`

\`better-sqlite3\` 12.9.0 ships a prebuilt binary for Node ABI 141, but Electron 41 bundles Node 24 (ABI 145). Every \`db:*\` IPC call fails with \`ERR_DLOPEN_FAILED\`.

## Fix

Adds \`@electron/rebuild\` as a devDep and a \`postinstall\` hook that rebuilds \`better-sqlite3\` against Electron's bundled Node version.

## Test plan
- [x] \`yarn install\` triggers postinstall and produces an arm64-145 binary
- [x] \`yarn dev:desktop\` starts cleanly — no DLOPEN errors in log

### DIFF
--- a/desktop/package.json
+++ b/desktop/package.json
@@ -6,7 +6,8 @@
   "scripts": {
     "dev": "electron-vite dev",
     "build": "electron-vite build",
-    "typecheck": "tsc --noEmit"
+    "typecheck": "tsc --noEmit",
+    "postinstall": "electron-rebuild -f -w better-sqlite3"
   },
   "dependencies": {
     "better-sqlite3": "^12.9.0",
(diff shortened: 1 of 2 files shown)


yagudaev:
Superseded — the same better-sqlite3 rebuild fix landed via #155.

PR #174, opened 2026-04-23
refactor(tracing-ui): migrate reader from raw SQL to Prisma

yagudaev:
## Why

Raw SQL in `tracing-ui/src/lib/db.ts` was fine for the scaffold, but we already know where this is going: once the collector grows beyond a single-host SQLite file we want to point the reader at Postgres without rewriting every query. Prisma gets us typed queries now and a clean migration path later. This PR does the reader migration and owns the connection via a driver adapter so the Prisma config stops leaking into env vars.

## How

- Generated a Prisma schema that mirrors the collector's DDL (traces / observations / sessions / media). The schema is read-only from this package — the collector owns migrations.
- Adopted `@prisma/adapter-better-sqlite3` so the runtime connection is constructed in `src/lib/db.ts` with `readonly: true`, `fileMustExist: false`, and a 5s busy timeout. No more `VOICECLAW_TRACING_DB_URL` indirection — we take the same `VOICECLAW_TRACING_DB` path the collector reads, with `~/.voiceclaw/tracing.db` as a fallback. The `url` on the datasource block in `schema.prisma` is now a placeholder the adapter never consults.
- Most reads stay on `$queryRaw` + `CAST(start_time_ns AS REAL)` for the reason documented below; the aggregation in `listSessions` also remains raw because Prisma's fluent API can't express the trace-level CTE without an N+1.

## Tradeoffs

- **BigInt on SQLite INTEGER columns is still the blocker.** I verified with a throwaway test that the driver adapter does *not* sidestep the engine's INT32 range check — the adapter inspects the DDL decltype via `stmt.columns()`, maps `INTEGER` → `ColumnTypeEnum.Int32`, and the engine rejects the string-encoded i64 value before we see it. This fails identically for fluent reads, `$queryRaw` without CAST, and even `select` projections of a single BigInt column. The only real fix is migrating the collector's DDL to `BIGINT`, which is out of scope for this PR. Hence every timestamp-bearing query projects `CAST(... AS REAL)` so the engine sees `Double`; the precision loss is sub-microsecond (well below the ms resolution the UI renders).
- We add `@prisma/adapter-better-sqlite3` (runtime) and keep `better-sqlite3` in `dependencies` because the adapter requires it as a runtime peer. `@types/better-sqlite3` lands in devDependencies because `better-sqlite3` ships no types and the adapter's `.d.ts` re-exports its `Options` type.
- `schema.prisma` carries a placeholder `url` on the datasource block. Prisma 6's schema validator still requires one at `prisma generate` time even though the adapter owns the runtime connection; Prisma 7 lifts this.

## Verification

- `yarn tsc --noEmit` clean
- `yarn build` clean (4 routes compile, no warnings)
- `yarn dev` + curl of `/`, `/sessions`, `/sessions/<real-id>`, `/users`, `/dashboard` — all 200, real session rows render, no server errors

### DIFF
--- a/tracing-ui/.env.example
+++ b/tracing-ui/.env.example
@@ -1,0 +1,4 @@
+# Prisma datasource for the tracing-ui reader. Points at the SQLite file the
+# tracing-collector writes. Default matches the collector's default path; set
+# this explicitly if you moved the DB or run against a snapshot.
+VOICECLAW_TRACING_DB_URL="file:/Users/you/.voiceclaw/tracing.db"
--- a/tracing-ui/.gitignore
+++ b/tracing-ui/.gitignore
@@ -2,3 +2,5 @@
 out/
 next-env.d.ts
 *.tsbuildinfo
+.env
+.env.local
--- a/tracing-ui/next.config.ts
+++ b/tracing-ui/next.config.ts
@@ -2,8 +2,9 @@ import { dirname } from "node:path"
 import { fileURLToPath } from "node:url"
 import type { NextConfig } from "next"
 
-// better-sqlite3 is a native module. Next.js's server-components bundler needs
-// to treat it as external so it isn't re-bundled for the server runtime.
+// @prisma/client ships native engine binaries; Next.js's server-components
+// bundler needs to treat it as external so the engines aren't re-bundled for
+// the server runtime (and so runtime file resolution works).
 //
 // `outputFileTracingRoot` is pinned to this workspace to silence Next's
 // "multiple lockfiles" warning when the repo is opened inside a broader yarn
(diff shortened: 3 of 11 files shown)


yagudaev:
Closing per discussion: the driver adapter experiment confirmed that `@prisma/adapter-better-sqlite3` doesn't bypass Prisma's INT32 range check either — the adapter's `mapDeclType` reads the collector's `INTEGER` DDL, maps to `ColumnTypeEnum.Int32`, and the engine enforces i32 range on fluent, `$queryRaw`, and single-column projections alike. The only actual fix would be migrating the collector schema from `INTEGER` to `BIGINT`, which is out of scope here.

Net value with the adapter was small: `readonly: true` on the reader + a fix for the `resolveDbUrl()` env-var bug. Not worth the `@prisma/client` + schema-duplication cost when we're not migrating to Postgres near-term.

Revisit when we actually move off SQLite — at that point Prisma's fluent API will work natively on the new backend and the migration path will earn its keep.

PR #177, opened 2026-04-23
feat(tracing-ui): turns timeline + media capture + UX polish

yagudaev:
## Why

The Latency tab was mostly empty columns (no STT/TTS for realtime models). The Sessions-first nav buried the Dashboard. Tool calls rendered no input/output. There was no way to replay the actual audio of a turn — the #1 request when debugging a bad response. This PR addresses all of that and builds the infrastructure so future media-rich debugging is additive.

## What landed

**Latency tab cleanup**
- Dropped `Endpointing`, `Voice (TTS)`, `Transcriber (STT)`, `To-Transport` — realtime models don't expose those stages and the columns were always empty.
- Renamed `LLM` to `Realtime`; kept `Brain (async)`; added a catch-all `Other` so unknown spans stay visible without being miscategorized as TTS.
- Added a **TTFT** column next to Total with inline tooltip: _"Time to first token. For audio, the first audio frame. For video, the first video frame. For text, the first text chunk."_ Derived from `gen_ai.response.first_token_ns` on the voice-turn span, or the earliest `first_token`/`first_chunk`/`first_audio_frame` event on any child span. Renders `—` when absent (see Deferred).

**Nav reorder**
- `Dashboard` first, then `Sessions`, then `Users`. `/` redirects to `/dashboard`.

**New Turns tab** (between Transcript and Logs)
- Left rail: clickable turn rows showing #, offset, duration.
- Middle: foldable **timeline** tree (indented by `parent_span_id`) with duration bars + relative start offsets. `voice-turn` is the root; `ask_brain`, `openclaw.chat_completions`, and `openclaw.llm` children nest under.
- Right: **Span Details** panel — name, duration, service, status, model, tokens (in/out/cached), cost, tool input/output, events, all attributes pretty-printed.
- URL state: `?turn=<traceId>&step=<spanId>` so reloads preserve view and reviewers can deep-link.

**Tool call details**
- Detail panel reads `gen_ai.tool.input` / `gen_ai.tool.output` with a legacy `langfuse.observation.input` / `langfuse.observation.output` fallback. JSON is pretty-printed; text rendered verbatim. Once the openclaw fork emits the `gen_ai.*` keys this panel will light up automatically.

**Full media capture pipeline** (opt-in: `VOICECLAW_MEDIA_CAPTURE=enabled`)
- Relay tees inbound mic PCM + outbound model PCM to `~/.voiceclaw/media/<sessionKey>/{user,assistant}-<turnId>.pcm` with a sidecar `{sampleRate, channels}` JSON; video frames land as numbered JPEGs under `video-<turnId>/` with `timings.json`.
- Relay stamps vendor-neutral `media.*` attrs on the voice-turn span: `media.user_audio.path|duration_ms|sample_rate|codec|bytes|provider` and the same for `assistant_audio`; `media.user_video.path|frame_count|duration_ms|codec|provider`.
- Collector indexes into the existing `media` SQLite table (added `UNIQUE(span_id, kind)` for idempotent upsert + a `trace_id` index).
- Tracing UI serves bytes from new `/api/media/[sessionId]/[...path]` route; PCM is re-wrapped to WAV on the fly for native browser playback. Path-traversal is blocked by anchoring to the session root.
- Turns tab renders user + assistant `<audio>` tracks with native controls, plus a canvas-based JPEG-sequence stepper (play/pause/scrub) for Gemini Live video frames.

**Vendor-neutral attribute prefixes**
- New emission uses `media.*` / `gen_ai.*` only. Existing `langfuse.observation.*` reads stay for back-compat; no new `langfuse.*` attrs introduced.

**Docs**
- New Starlight page at `docs/src/content/docs/tracing-ui/turns-and-media.md` covering the Turns tab, the `media.*` attr family, the API route, and what's deferred.

## Deferred (follow-ups)

- **Langfuse Media upload** — flat-file capture works today; `media.<role>.provider` is `local`. Wiring `@langfuse/media` uploads and populating `media.<role>.url` / `media.<role>.langfuse_id` is a separate PR so this one stays reviewable.
- **Relay emission of `gen_ai.response.first_token_ns`** — UI is ready; the relay still needs to stamp the attribute (or a `first_token` event) on the voice-turn / realtime-LLM span. Until then the TTFT column shows `—`.

## Test plan

<details><summary>Expand</summary>

- [x] `yarn tsc --noEmit` clean in `tracing-ui`, `tracing-collector`, `relay-server`
- [x] `yarn build` clean in `tracing-ui` (Next 15.5.15, including the new `/api/media/[sessionId]/[...path]` route)
- [x] `relay-server/test/test-media-capture.ts` — flat-file writes, sidecar, per-turn isolation, disabled-mode no-op
- [x] `tracing-collector/test/test-media-schema.ts` — `UNIQUE(span_id, kind)` upsert semantics + `COALESCE` partial-update
- [x] `tracing-collector/test/test-media-ingest.ts` — OTLP-JSON ingest with `media.*` attrs produces the expected rows; re-ingest is idempotent
- [x] `tracing-ui/test/test-turns-reader.ts` — `listMediaForSession` + `listObservationsForTrace` return `events_json` and the new tool input/output keys
- [x] `relay-server/test/test-brain-traceparent.ts` still passes
- [x] `relay-server/test/test-gemini-unit.ts` still passes
- [ ] Manual: open a voice call with `VOICECLAW_MEDIA_CAPTURE=enabled`, verify PCM files on disk + attrs on span + Turns tab renders audio players that play back the real audio
- [ ] Manual: screen-share into a call, verify `video-<turnId>/*.jpeg` + `timings.json` land, Turns tab plays the frame sequence
- [ ] Manual: click a `voice-turn` → `ask_brain` → `openclaw.chat_completions` timeline step, verify span details render with tokens/cost/attributes

</details>

### DIFF
--- a/relay-server/src/media/capture.ts
+++ b/relay-server/src/media/capture.ts
@@ -1,0 +1,323 @@
+// Per-session media capture: tees client-mic PCM + model-output PCM to on-disk
+// flat files so the tracing UI can replay each turn. Video frames (Gemini Live
+// screen share) land as numbered JPEGs with a sidecar timings.json.
+//
+// Files live under `~/.voiceclaw/media/<sessionKey>/`:
+//
+//   user-<turnId>.pcm            mono PCM16 little-endian
+//   user-<turnId>.pcm.json       { "sampleRate": 24000, "channels": 1 }
+//   assistant-<turnId>.pcm
+//   assistant-<turnId>.pcm.json
+//   video-<turnId>/0000.jpeg     (one JPEG per frame)
+//   video-<turnId>/timings.json  { "frames": [{ "offset_ms": 0, "file": "0000.jpeg" }, …] }
+//
+// Capture is OPT-IN via `VOICECLAW_MEDIA_CAPTURE=enabled` (default off) so it
+// never surprises prod operators with disk usage. When enabled, the session
+// owns one instance; the relay calls `onUserAudioChunk` / `onAssistantAudioChunk`
+// / `onVideoFrame` during the turn and `finalizeTurn()` at turn.ended to
+// return the sidecar attrs to stamp on the voice-turn span.
+//
+// Langfuse Media upload is deferred behind a `enableLangfuseUpload` flag that
+// is currently hardcoded to false — the Langfuse v5 Media API requires an
+// external SDK call that we'd rather wire under a separate PR after this
+// flat-file capture has shaken out in the real session pipeline. Local path
+// capture alone is a huge UX win already.
+
+import { createHash } from "node:crypto"
+import { createWriteStream, existsSync, mkdirSync, type WriteStream } from "node:fs"
+import { promises as fs } from "node:fs"
+import { homedir } from "node:os"
+import { join, resolve } from "node:path"
+import { log } from "../log.js"
+
+export type MediaCaptureAttrs = {
+  // Stamped on the voice-turn span. Vendor-neutral `media.*` prefix.
+  [key: string]: string | number | boolean | undefined
+}
+
+export type MediaCaptureConfig = {
+  // Defaults: `VOICECLAW_MEDIA_CAPTURE=enabled` + `~/.voiceclaw/media`.
+  enabled?: boolean
+  rootDir?: string
+  userSampleRate?: number
+  assistantSampleRate?: number
+}
+
+export class MediaCapture {
+  private config: Required<Omit<MediaCaptureConfig, "enabled">> & { enabled: boolean }
+  private sessionDir: string | null = null
+  private sessionKey: string | null = null
+  private currentTurn: TurnState | null = null
+
+  constructor(config: MediaCaptureConfig = {}) {
+    const enabled = config.enabled ?? (process.env.VOICECLAW_MEDIA_CAPTURE === "enabled")
+    this.config = {
+      enabled,
+      rootDir:
+        config.rootDir
+        ?? process.env.VOICECLAW_MEDIA_DIR
+        ?? join(homedir(), ".voiceclaw", "media"),
+      // 24 kHz is VoiceClaw's client mic default and Gemini's output rate.
+      userSampleRate: config.userSampleRate ?? 24000,
+      assistantSampleRate: config.assistantSampleRate ?? 24000,
+    }
+  }
+
+  startSession(sessionKey: string): void {
+    if (!this.config.enabled) return
+    // Sanitize — session keys can include `/` and other path chars when callers
+    // derive them from external IDs. Hash anything unsafe so paths stay flat
+    // and predictable.
+    const safe = isSafeSegment(sessionKey) ? sessionKey : hashShort(sessionKey)
+    this.sessionKey = safe
+    this.sessionDir = join(this.config.rootDir, safe)
+    try {
+      mkdirSync(this.sessionDir, { recursive: true })
+    } catch {
+      // If we can't create the dir, disable capture for this session rather
+      // than crash the relay.
+      this.sessionKey = null
+      this.sessionDir = null
+    }
+  }
+
+  startTurn(turnId: string): void {
+    if (!this.config.enabled || !this.sessionDir) return
+    // Close any leftover turn state — shouldn't normally fire but guards
+    // against a turn.started without a preceding turn.ended.
+    void this.finalizeTurn()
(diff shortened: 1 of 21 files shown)


relay-server/src/media/capture.ts
```diff
@@ -0,0 +1,299 @@
+// Per-session media capture: tees client-mic PCM + model-output PCM to on-disk
```
yagudaev, line 1: does the media capture belong here or on the `tracing-collector` side?

yagudaev: **Consensus: belongs on the relay side (where it is today).** I asked Codex + Gemini to sanity-check.

**My take**: The bytes already live on the relay. PCM16 @ 24 kHz × 2 streams = ~96 KB/s per session. Shipping that to the collector means pushing bulk binary over OTLP (which isn't designed for it — span events cap small, base64 blows size past exporter limits) just to have the collector write it back to disk. Relay captures, collector indexes metadata. Clean layering.

**Codex agreed** — same verdict (A), same reasoning, with one refinement worth baking in:

> "If you later need cross-host access, add an async upload/storage layer behind the relay and emit `media.*.url`/IDs, not raw bytes to the collector."

> "A + async media sink: relay writes locally first for low-latency durability, then a background uploader promotes files to object storage / Langfuse Media and updates metadata. Collector still only stores references."

> "B introduces network backpressure, retries, partial delivery, ordering problems, and 'collector unavailable' failure modes into media capture. That is the wrong dependency direction."

Codex also flagged one worthwhile caveat: **make backpressure on the relay's local writes explicit** — at these rates it's manageable, but unbounded buffering on slow disks is the main local risk. I'll add an error handler + a max-queue cap as part of the review-finding cleanup I'm already doing on this PR (dual-reviewer also caught this — `createWriteStream` has no `.on('error')` today).

**Gemini**: CLI hung for >5 min on the same prompt size that choked it on the PR #177 diff review earlier. Skipping its take — will add if it returns.

### Practical implications for this PR

- Keep capture here. ✅
- Wire async Langfuse Media upload in a follow-up PR (already in the Deferred list).
- Add stream error handlers + a size guard (folding into the critical-fixes commit I'm pushing next).
- When we split relay + collector onto different hosts, we'll rely on `media.*.url` (Langfuse + S3 / object storage) rather than local paths. Until then the local fast-path wins.

Does this answer it? If you'd rather extract `relay-server/src/media/` into a shared `packages/media-capture` workspace so a future web-client relay can reuse the code, happy to do that in a follow-up too.

yagudaev: Update — Gemini returned too. Also verdict **A**. Agrees on all the same points (OTLP not for bulk, backpressure, cross-host scalability, separation of concerns).

One angle Gemini raises that Codex didn't, worth chewing on:

> "Third option (Hybrid): Stick with A, but ensure the Relay provides a **signed URL or local file endpoint** via its metadata. ... This keeps the Tracing UI decoupled — it reads the path from the SQLite database (via the Collector) but fetches the actual bytes from the Relay (or S3)."

Right now the Tracing UI reads media bytes **directly from its own filesystem** (`~/.voiceclaw/media/`). That's fine when UI + relay + collector are all on one machine. When we split them in prod, the UI would need to either:

- Mount shared storage (NFS, sync daemon — brittle), or
- Fetch via an HTTP endpoint on the relay (Gemini's suggestion), or
- Fetch via object storage URL (`media.*.url` — already the plan for Langfuse Media)

The cleanest path is probably option 3 (object storage URL everywhere) with option 2 as a dev-only fallback. Filing as a follow-up — not fixing in this PR.

Gemini's two other specific phrases worth capturing:

> "OTLP is designed for telemetry signals (logs, metrics, traces), not raw binary streams. Sending multi-megabyte media payloads through an OTel exporter is an anti-pattern."

> "Relay: Source of Truth for the Session (Media + State). Collector: Source of Truth for the Trace (Metadata + Timing)."

Good mental model to keep.

tracing-ui/src/components/TurnsTab.tsx
```diff
@@ -0,0 +1,726 @@
...
+          onChange={(e) => setFrameIdx(Number(e.target.value))}
+          className="flex-1"
+        />
+        <span className="tabular-nums text-zinc-500">
+          {frameIdx + 1}/{frameCount}
+        </span>
+      </div>
+    </div>
+  )
+}
```
yagudaev, line 566: nit: might be worth while moving some of these components to their own files file is getting long

PR #190, opened 2026-04-24
Add Privacy Policy, Terms of Service, and /legal pages

yagudaev:
## Why

Unblocks two downstream pieces:
1. **Google OAuth consent screen verification** requires public Privacy Policy + Terms URLs before Google approves the app for production (removes the "unverified app" scary screen).
2. **Apple App Store Review Guideline 1.2** requires UGC moderation clauses in our Terms of Service before they'll accept the mobile app.

Google's verification review is 1–6 weeks, so standing these pages up now puts that clock on the right side of the Phase 1 identity PR.

## What's in

| Route | Page | Size |
|---|---|---|
| \`/privacy\` | Privacy Policy | ~310 lines of structured TSX |
| \`/terms\` | Terms of Service | ~250 lines of structured TSX |
| \`/legal\` | Index linking both | ~40 lines |

Plus a shared \`LegalPage\` component (\`website/src/components/legal/LegalPage.tsx\`) with minimal styling — back-to-home link, title, lede, body prose, footer with Nano 3 Labs Ltd. address + support email.

## Voice / shape

- Parallels audiowaveai.com's structure (same company, same Vancouver address, same BC governing law)
- **Shorter and more human** than the full 28-section auto-generated template — VoiceClaw doesn't have payments, IAP, or complex account logic yet, so we only describe what's actually true
- **Privacy "short version" at the top** — what we store, what we don't, what we use, what you can do. Plain language.
- **UGC clauses in Terms** lean on "we're a frontend for providers you configured" to shift content responsibility to Google/OpenAI/xAI where Apple already trusts their moderation (the shield called out in NAN-662)
- **Explicitly calls out what we don't receive:** audio, transcripts, API keys, conversations. This is the actual differentiator.

## Minimal styling

The pages use the existing shadcn dark-mode Tailwind tokens. Codex is redesigning the marketing site to match the Option B brand (paper + Fraunces + warm muted text); this component is deliberately neutral so it slots in cleanly. No custom colors, no custom fonts — just readable semantic markup.

### /privacy
Privacy

### /terms
Terms

(Screenshots available in the branch at \`legal-privacy.png\` / \`legal-terms.png\` via Playwright; adding them via comment below)

## Test plan

- [x] \`yarn build\` — all three pages prerender as static (\`○ /legal\`, \`○ /privacy\`, \`○ /terms\`)
- [x] Visual check at localhost:3100 — pages render, links work, typography reads
[...]

PR #191, opened 2026-04-24
Phase 1 identity: Prisma + Arctic + jose + OAuth routes (env-gated)

yagudaev:
## Why

Full scaffolding for the browser → desktop/mobile OAuth handoff. Every route is env-gated so the PR ships green without creds; Michael drops credentials in Vercel env vars once Google Cloud + Apple Developer clients are provisioned. This unblocks Phase 5 (wizard wired to real sign-in).

### Prisma schema
\`\`\`
User          — email, name, googleSub, appleSub, appleIdEmail,
                mobileAccessRequestedAt/grantedAt, analyticsOptedOutAt
Device        — tokenHash (SHA-256), label, platform, lastSeenAt, revokedAt
AuthTicket    — ticketHash, expiresAt (~90s), consumedAt, targetPlatform
\`\`\`

### Auth utilities (\`website/src/lib/auth/\`)
- \`db.ts\` — Prisma singleton (HMR-safe)
- \`providers.ts\` — Arctic Google + Apple config, lazy-loaded
- \`tokens.ts\` — ticket mint/hash, JWT device token (HS256, 1y), SHA-256 DB storage
- \`cookies.ts\` — HttpOnly state + PKCE verifier cookies
- \`deep-link.ts\` — builds \`voiceclaw://auth/callback?ticket=...\` + fallback HTML

### Routes
| Method | Path | Purpose |
|---|---|---|
| GET | \`/api/auth/google/start\` | State + PKCE, redirect to Google |
| GET | \`/api/auth/google/callback\` | Exchange, upsert, mint ticket, deep-link bounce |
| GET | \`/api/auth/apple/start\` | Same for Apple web flow |
| GET | \`/api/auth/apple/callback\` | Apple-specific gotchas handled |
| POST | \`/api/auth/ticket/redeem\` | App exchanges ticket → device token |
| GET | \`/api/me\` | Bearer token → user + device + mobile access state |
| GET | \`/auth/failed\` | Minimal error page with reason code |

### Env contract
\`.env.example\` documents every var. Every route returns \`501 { error: \"not_configured\" }\` when its required vars are missing, so the site still boots without creds.

### Build wiring
- \`prisma generate\` in \`postinstall\` + \`build\`
- \`db:push\` + \`db:studio\` for dev
- Pinned to **Prisma 6** (stable API; v7 moved datasource config to a new \`prisma.config.ts\` that would be premature churn)

## What's NOT in this PR

- Real Google Cloud OAuth credentials (you provision; drop in Vercel env)
- Real Apple Sign In service ID + .p8 (you provision)
- Real \`DATABASE_URL\` → Neon (you provision the schema on existing Neon)
- iOS-native Apple sign-in route (\`POST /api/auth/apple/native\`) — arrives with mobile PR
- Desktop \`voiceclaw://\` URL-scheme handler — arrives with wizard wiring PR
- Settings UI for revoking devices — later PR

## Security I want extra eyes on

- **Ticket scheme**: SHA-256(base64url(32 random bytes)) — 256 bits entropy, one-shot, ~90s TTL. Shape OK?
- **Device token**: JWT (HS256) + SHA-256 of whole token stored in DB. Both must match at \`/me\`. Revoke = flip \`revokedAt\`. JWT stays verifiable but hash lookup misses → 410. Reasonable?
- **State cookie**: \`\${state}.\${target}\` suffix so callback knows deep-link scheme. Target hint isn't signed (it's UX, not security). Fine?
- **Apple PEM**: stripped + base64-decoded to DER in \`providers.ts\`. Never logged. Alternative: \`.p8\` on disk read by path. Thoughts?

## Test plan

- [x] \`yarn build\` — 6 dynamic API routes + 1 static page compile cleanly
- [x] \`prisma generate\` runs in postinstall + build
- [x] \`DATABASE_URL=postgresql://dummy prisma validate\` passes
- [ ] Reviewer: walk the flow — start → callback → ticket → redeem → /me. Look for ticket-replay or orphaned-device races.
- [ ] Once creds wired: curl test against localhost

## Dependencies

- Stacks on **#190** (Privacy + Terms) — cleanly separate (different files)
- Blocks: wizard wiring PR (needs working /api/auth/ticket/redeem + /api/me)

### DIFF
--- a/website/.env.example
+++ b/website/.env.example
@@ -1,0 +1,50 @@
+# VoiceClaw website — identity layer environment variables.
+#
+# Copy this to `.env.local` for dev. In production, set these via
+# Vercel → Project → Settings → Environment Variables.
+#
+# Every route gracefully degrades to 501 "not_configured" when its
+# required vars are unset, so this file can ship empty and the app
+# still boots.
+
+# ---------------------------------------------------------------------------
+# Database (Neon / Postgres)
+# ---------------------------------------------------------------------------
+# Full connection string including ?sslmode=require for Neon.
+# e.g. postgresql://user:pass@ep-xxx.us-east-1.aws.neon.tech/voiceclaw?sslmode=require
+DATABASE_URL=
+
+# ---------------------------------------------------------------------------
+# Token signing
+# ---------------------------------------------------------------------------
+# Generate with: openssl rand -hex 48
+# Rotating this invalidates every issued device token.
+AUTH_JWT_SECRET=
+
+# ---------------------------------------------------------------------------
+# Google OAuth
+# ---------------------------------------------------------------------------
+# Create a Web application OAuth client at
+#   https://console.cloud.google.com/apis/credentials
+# Authorized redirect URI: <NEXT_PUBLIC_APP_URL>/api/auth/google/callback
+GOOGLE_CLIENT_ID=
+GOOGLE_CLIENT_SECRET=
+
+# ---------------------------------------------------------------------------
+# Apple Sign In
+# ---------------------------------------------------------------------------
+# Service ID, Team ID, Key ID, and the contents of the .p8 private key
+# file from Apple Developer > Certificates, Identifiers & Profiles > Keys.
+# APPLE_PRIVATE_KEY should be the full PEM including the BEGIN/END lines.
+APPLE_CLIENT_ID=
+APPLE_TEAM_ID=
+APPLE_KEY_ID=
+APPLE_PRIVATE_KEY=
+
+# ---------------------------------------------------------------------------
+# Public URL of this deployment
+# ---------------------------------------------------------------------------
+# Used to build OAuth redirect URIs. Vercel populates VERCEL_URL
+# automatically for preview deployments; set NEXT_PUBLIC_APP_URL
+# explicitly for production so the prod URL is stable.
+NEXT_PUBLIC_APP_URL=http://localhost:3000
--- a/website/.gitignore
+++ b/website/.gitignore
@@ -32,6 +32,8 @@ yarn-error.log*
 
 # env files (can opt-in for committing if needed)
 .env*
+# keep the example file so devs know what to populate
+!.env.example
 
 # vercel
 .vercel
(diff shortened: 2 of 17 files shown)


yagudaev:
Rebased onto main (resolved a yarn.lock conflict by regenerating from the current lock). Ran Codex + Gemini reviews.

**Addressed (commit 5c23f78):**
- **Ticket redeem race (`ticket/redeem/route.ts`)** — replaced find-then-update with atomic `updateMany` keyed on `consumedAt: null`. Two concurrent redeems now race exactly once at the DB level; the loser returns 410 instead of minting a duplicate device.
- **Public-interface ordering (`tokens.ts`)** — moved `isSigningKeyConfigured` above the `// Helpers` banner. Matches the repo rule (public interface at top, helpers at bottom).
- **No-op ternary (`deep-link.ts`)** — `params.target === "mobile" ? "voiceclaw" : "voiceclaw"` removed; expanded the comment to explain why the scheme is uniform (dev/staging variants are routed app-side).
- **Redundant index (`schema.prisma`)** — dropped `@@index([email])`. `@unique` already creates a b-tree index on the column.

**Deferred with reasoning:**
- **Same-email collision between Google and Apple sign-ups** (Codex). Real edge case, but fixing it means choosing an account-linking policy (auto-merge on verified email? block with a helpful error? require explicit "link your Apple ID" flow?). That's a product decision, not a bug fix — leaving it for a follow-up ticket once we have a UX for it. The current behavior (unique-constraint error bubbles up) is loud, not silent, so no user will end up with a silently duplicated account.
- **ID token signature verification** (Codex). `arctic.decodeIdToken` trusts the fact that the token came directly from Google's / Apple's token endpoint over TLS (the RFC-8725 token-endpoint-trust pattern). Adding JWKS verification is defense-in-depth and a reasonable follow-up, but not a correctness bug.

**False positives (Gemini):** `Bearers+` regex (actual code is `Bearer\s+`), "space in import path" / "space in mailto" (Gemini mis-parsed its own stat-error output), `next@16.2.4` version (correct — Next 16 is current). No action needed.

Build + typecheck verified on the rebased branch.

PR #214, opened 2026-04-25
feat(desktop): floating call bar

yagudaev:
## Summary

Granola-style floating window that appears during realtime sessions: small dark pill, brand mark on top, animated audio waveform in the middle, drag handle on hover. Always-on-top, frameless, transparent, `focusable: false` so it doesn't steal focus from the app the user is actively in.

## What's wired

- `desktop/src/main/call-bar.ts` — window lifecycle (88×240 px bottom-right pill), screen-saver-level always-on-top, position persisted in SQLite `settings` table, 30 Hz audio-level broadcast, right-click context menu (mute / end call / hide bar).
- `desktop/src/renderer/src/call-bar/CallBar.tsx` + `call-bar.css` — reactive pip animation. Colors follow the speaker (`is-user-speaking` → `--brand-accent`, `is-ai-speaking` → `--brand-paper`) with a subtle idle shimmer between turns. Token block and body rules are scoped to `body.call-bar-view` so the stylesheet doesn't leak when both bundles share a renderer.
- Hooked into `setCallActive` from #213 — bar appears when a call connects, fades out when it ends.
- `SettingsPage` toggle to disable.
- Starlight docs page at `docs/src/content/docs/desktop/call-bar.mdx`.

## Renderer architecture

Single renderer bundle. Both the main window and the call-bar window load `desktop/src/renderer/index.html`; the call-bar BrowserWindow loads it with `?view=call-bar`. `src/renderer/src/main.tsx` reads the query string and dynamically imports either `App` + `index.css` or `CallBar` + `call-bar.css`, keeping each window's initial payload minimal. We tried a multi-entry `rollupOptions.input` first; electron-vite silently drops the second entry, so the query-string approach is simpler and avoids fighting the framework.

## Test plan

- [ ] `yarn build` — one `out/renderer/index.html`, `CallBar` and `call-bar.css` emitted as code-split chunks.
- [ ] `yarn typecheck` clean.
- [ ] `yarn dev` — start a call, pill appears bottom-right with the brand mark and animated pips.
- [ ] Drag the pill, quit the app, relaunch — pill reappears at the saved position.
- [ ] Right-click the pill — context menu shows Mute mic / End call / Hide bar.
- [ ] Toggle off in Settings — bar no longer appears on the next call.


desktop/src/renderer/src/call-bar/call-bar.css
```diff
@@ -0,0 +1,214 @@
...
+/* Focusable hit-target for the mark click-to-focus-main-window */
+.call-bar__focus-zone {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  height: 28px;
+  -webkit-app-region: no-drag;
+  cursor: pointer;
+}
```
yagudaev, line 214: TODO: lets convert this to tailwind css, we might need minimum css for just the animation bit

desktop/src/renderer/src/main.tsx
```diff
@@ -1,14 +1,42 @@
...
+      import('./call-bar/CallBar'),
+      import('./call-bar/call-bar.css'),
+    ])
+    root.render(
+      <StrictMode>
+        <CallBar />
+      </StrictMode>,
+    )
+    return
+  }
```
yagudaev, line 28: Is there a cleaner way to do this?

PR #215, opened 2026-04-25
feat(desktop): screen-share indicator (capture frame + call-bar pip)

yagudaev:
## Why

macOS already shows an orange dot in the menu bar when the screen is being recorded. That dot **disappears in fullscreen apps** — Keynote, fullscreen video, immersive demos — exactly the moment a "you are live" signal matters most. This PR adds two app-owned indicators that ride above fullscreen surfaces so the user always knows which display is going out the wire.

## Two surfaces, one signal

1. **Capture frame.** A transparent, click-through, always-on-top `BrowserWindow` sized to the full bounds of the shared display. Draws a 3 px rust border around the perimeter and a small "Sharing this screen" pill at the top center. Border breathes slowly (3 s pulse, 100 % to 55 %) so it stays present without becoming alarmist.
2. **Call-bar pip.** A small solid rust dot in the top-right of the brand mark on the floating call bar — visible only while a share is active.

Both come from one event: the chat renderer fires `screenShare.setActive({ active, displayId })` on share start (and on `track.onended`, our stop button, and call end). Main shows / hides the frame on the chosen display and fans the same state into the call-bar window for the pip.

## Architecture

- New `desktop/src/main/screen-frame.ts` — single `BrowserWindow` reused across share sessions. `frame:false`, `transparent:true`, `focusable:false`, `setIgnoreMouseEvents(true, { forward:false })`, `setAlwaysOnTop(true, 'screen-saver')`, `visibleOnFullScreen:true`, `enableLargerThanScreen:true`.
- Mirrors the call-bar's **single-bundle** pattern — same `index.html` loaded with `?view=screen-frame`, dynamic-imported in `main.tsx`. Verified in build output: `screen-frame-*.css` (3 kB) + `ScreenFrame-*.js` (1.2 kB) chunks emit alongside the call-bar's chunks; neither window pays for the other's CSS / React tree.
- IPC contract in `desktop/src/preload/index.ts` under `screenShare`: producer `setActive(payload)`, consumer `onState(handler)` for the call-bar, plus `ready()` for the frame's renderer.
- Settings toggle (`screen_frame_enabled`, default ON) lives in the existing **Call Bar** card.
- Docs: extended `docs/src/content/docs/desktop/call-bar.mdx` with a new "Screen-share indicators" section explaining the why, the IPC flow, and how to disable.

## V1 descopes

- The frame is **read-only** — clicking it does not stop the share (the whole window is `setIgnoreMouseEvents`, by design). Stop via the chat UI, the call-bar context menu, or macOS's own Stop Sharing affordance.
- Window sources (sharing a single app window) carry no `display_id`; main falls back to the primary display in that case.
- No "not being shared" indicator on the OTHER monitor (intentional — too noisy).

## Branching

This PR is based on `michael/nan-desktop-call-bar` (not main) because it depends on the call-bar's single-bundle `?view=` routing pattern. Will rebase onto main after PR #214 merges.

## Verify

- [x] `yarn typecheck` clean
- [x] `yarn build` clean — emits `screen-frame-*.css` (3.06 kB) and `ScreenFrame-*.js` (1.19 kB) chunks alongside the call-bar's chunks
- [x] `yarn dev` boots cleanly through main + preload + renderer + electron launch
- [x] Docs build clean (Starlight)
- [ ] Manual smoke on a multi-monitor Mac: start a share, frame appears on the captured display, rust pip appears on call-bar mark, both clear on stop / track.onended

### DIFF
--- a/desktop/src/main/index.ts
+++ b/desktop/src/main/index.ts
@@ -16,6 +16,7 @@ import {
   createCallBar,
   destroyCallBar,
   focusMainFromCallBar,
+  getCallBarWindow,
   hideCallBar,
   markCallBarReady,
   registerCallBarHooks,
@@ -23,6 +24,14 @@ import {
   showCallBar,
   showCallBarContextMenu,
 } from './call-bar'
+import {
+  createScreenFrame,
+  destroyScreenFrame,
+  getScreenFrameWindow,
+  hideScreenFrame,
+  markScreenFrameReady,
+  showScreenFrame,
+} from './screen-frame'
 import { serviceManager } from './services/service-manager'
 import { startBundledOpenClaw } from './services/openclaw-gateway'
 import { ensureDefault as ensureLaunchAtLoginDefault } from './login-items'
@@ -133,6 +142,58 @@ app.whenReady().then(async () => {
     markCallBarReady()
   })
 
+  // Screen-share indicator -------------------------------------------------
+  // Capture-frame BrowserWindow + the rust pip on the call-bar Mark are
+  // both driven by a single signal: when the chat renderer starts a
+  // share, it fires `screen-share:setActive`; when the share ends (user
+  // clicks Stop, track.onended, or session ends), it fires it again
+  // with active=false. The handler:
+  //   1. shows / hides the perimeter frame on the chosen display
+  //   2. forwards the same state to the call-bar window so its CSS pip
+  //      lights up live (no DB / settings round-trip needed)
+  createScreenFrame({ isDev, rendererUrl: process.env.ELECTRON_RENDERER_URL })
+
+  // Lifecycle signal owned by the screen-frame renderer only — reject
+  // anything coming from the main app or any other webContents so a
+  // compromised renderer can't flush a queued show on our behalf.
+  ipcMain.handle('screen-frame:ready', (event) => {
+    const frame = getScreenFrameWindow()
+    if (!frame || event.sender !== frame.webContents) return
+    markScreenFrameReady()
+  })
+
+  // Privileged: this handler controls an alwaysOnTop screen-saver-level
+  // BrowserWindow plus call-bar pip state. Only the main chat window
+  // is allowed to call it, and the payload shape must be exact.
+  ipcMain.handle(
+    'screen-share:setActive',
+    (event, payload: unknown) => {
+      const main = getMainWindow()
+      if (!main || event.sender !== main.webContents) return
+      if (!payload || typeof payload !== 'object') return
+
+      const p = payload as { active?: unknown; displayId?: unknown }
+      if (typeof p.active !== 'boolean') return
+      const displayId =
+        typeof p.displayId === 'number' && Number.isFinite(p.displayId)
+          ? p.displayId
+          : undefined
+
+      if (p.active && isScreenFrameEnabled()) {
+        showScreenFrame(displayId)
+      } else {
+        hideScreenFrame()
+      }
+
+      // Mirror state into the call-bar so its rust pip can react in
+      // real time. Use .send (not the queued visibility broadcaster)
+      // because the bar may be hidden — we still want it to remember
+      // the latest state the next time it shows.
+      const bar = getCallBarWindow()
+      bar?.webContents.send('call-bar:screen-share', { active: p.active })
+    },
+  )
+
   ipcMain.handle('call-bar:focus-main', () => {
     focusMainFromCallBar()
   })
@@ -209,6 +270,7 @@ app.on('before-quit', async (event) => {
 app.on('will-quit', () => {
   serviceManager.stopAll()
   destroyCallBar()
+  destroyScreenFrame()
   destroyTray()
   closeDb()
 })
@@ -252,6 +314,19 @@ function isCallBarEnabled(): boolean {
     const row = db
       .prepare('SELECT value FROM settings WHERE key = ?')
       .get('call_bar_enabled') as { value: string } | undefined
+    // Default ON — explicit opt-out only.
+    return row?.value !== 'false'
+  } catch {
+    return true
+  }
+}
+
+function isScreenFrameEnabled(): boolean {
+  try {
+    const db = getDb()
+    const row = db
+      .prepare('SELECT value FROM settings WHERE key = ?')
+      .get('screen_frame_enabled') as { value: string } | undefined
     // Default ON — explicit opt-out only.
     return row?.value !== 'false'
   } catch {
--- a/desktop/src/main/screen-frame.ts
+++ b/desktop/src/main/screen-frame.ts
@@ -1,0 +1,165 @@
+import { BrowserWindow, screen } from 'electron'
+import { join } from 'node:path'
+
+// Screen-share capture frame — a thin always-on-top, click-through
+// rust-colored border drawn around the perimeter of whichever display
+// the user is currently sharing. Mirrors the floating call-bar pattern:
+// one BrowserWindow created lazily, reused across share sessions, fully
+// passive (no focus, no input). Modeled on macOS's menu-bar "screen
+// recording" orange dot — except that one disappears in fullscreen apps
+// like Keynote / video / immersive demos, exactly when the indicator
+// matters most. This window rides above fullscreen surfaces so the user
+// can always see at a glance which monitor is going out the wire.
+//
+// The window is sized to the target display's full bounds (not workArea)
+// so the border traces the absolute edge of the screen. Electron clips
+// to workArea on alwaysOnTop normal windows, but with `screen-saver`
+// level + `enableLargerThanScreen` the frame extends to the full
+// display rectangle.
+
+let frameWindow: BrowserWindow | null = null
+let isReady = false
+let queuedShow: { displayId?: number } | null = null
+
+type CreateOptions = {
+  isDev: boolean
+  rendererUrl?: string
+}
+
+export function createScreenFrame(options: CreateOptions): BrowserWindow {
+  if (frameWindow && !frameWindow.isDestroyed()) return frameWindow
+
+  // Initial bounds don't matter — we re-set them on every show. Use the
+  // primary display so the window has somewhere sane to live until then.
+  const primary = screen.getPrimaryDisplay()
+  const { x, y, width, height } = primary.bounds
+
+  frameWindow = new BrowserWindow({
+    x,
+    y,
+    width,
+    height,
+    frame: false,
+    transparent: true,
+    resizable: false,
+    movable: false,
+    minimizable: false,
+    maximizable: false,
+    skipTaskbar: true,
+    alwaysOnTop: true,
+    hasShadow: false,
+    // Passive observer — the frame is purely visual. Never steals focus
+    // and never receives clicks.
+    focusable: false,
+    enableLargerThanScreen: true,
+    show: false,
+    webPreferences: {
+      preload: join(__dirname, '../preload/index.js'),
+      contextIsolation: true,
+      nodeIntegration: false,
+    },
+  })
+
+  // Ride above fullscreen surfaces (Keynote, fullscreen video, immersive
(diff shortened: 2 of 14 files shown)


desktop/src/main/index.ts
```diff
@@ -135,6 +143,43 @@ app.whenReady().then(async () => {
...
+  // share, it fires `screen-share:setActive`; when the share ends (user
+  // clicks Stop, track.onended, or session ends), it fires it again
+  // with active=false. The handler:
+  //   1. shows / hides the perimeter frame on the chosen display
+  //   2. forwards the same state to the call-bar window so its CSS pip
+  //      lights up live (no DB / settings round-trip needed)
+  createScreenFrame({ isDev, rendererUrl: process.env.ELECTRON_RENDERER_URL })
+
+  ipcMain.handle('screen-frame:ready', () => {
+    markScreenFrameReady()
```
yagudaev, line 162: **[Codex]** **Major** — `screen-frame:ready` accepts any renderer process, so the main app or any compromised renderer can mark the hidden frame ready and flush a queued show even though only the screen-frame window should own this lifecycle signal. Validate `event.sender` against the screen-frame BrowserWindow webContents before calling `markScreenFrameReady()`, or expose a scoped preload only in the screen-frame window.

desktop/src/main/index.ts
```diff
@@ -135,6 +143,43 @@ app.whenReady().then(async () => {
...
+    (_e, payload: { active: boolean; displayId?: number }) => {
+      const active = Boolean(payload?.active)
+      const displayId =
+        typeof payload?.displayId === 'number' ? payload.displayId : undefined
+
+      if (active && isScreenFrameEnabled()) {
+        showScreenFrame(displayId)
+      } else {
+        hideScreenFrame()
+      }
```
yagudaev, line 186: **[Codex]** **Major** — `screen-share:setActive` trusts any renderer that can access the preload API to show or hide an always-on-top `screen-saver` level BrowserWindow and fan state into the call bar. Validate `event.sender` as the main chat window, validate the payload shape explicitly, and reject non-finite or unexpected `displayId` values before changing privileged window state.

desktop/src/renderer/src/lib/screen-capture.ts
```diff
@@ -42,6 +51,16 @@ export class ScreenCapture {
       } as any,
     })
 
+    // When the user clicks macOS's "Stop Sharing" affordance (or the
+    // OS otherwise revokes the share), the video track fires `ended`.
+    // Bubble that up so callers can clear UI state and turn off the
+    // screen-share indicator.
+    if (onEnded) {
+      this.stream.getVideoTracks().forEach((track) => {
+        track.addEventListener('ended', onEnded, { once: true })
```
yagudaev, line 83: **[Codex]** **Minor** — The `ended` listener added to each video track is never removed if capture stops normally via `ScreenCapture.stop()`. `{ once: true }` mostly mitigates this (the listener self-removes on fire), but on a normal stop the listener never fires and only releases when the track is GC'd, so callbacks holding closures over a stale page lifecycle can hang on briefly. Store the wrapped listeners and `removeEventListener` them in `stop()` before stopping the tracks.

desktop/src/main/screen-frame.ts
```diff
@@ -0,0 +1,170 @@
...
+
+function resolveDisplay(displayId?: number) {
+  if (typeof displayId === 'number') {
+    const found = screen.getAllDisplays().find((d) => d.id === displayId)
+    if (found) return found
+  }
+  return screen.getPrimaryDisplay()
+}
+
+export function getCurrentScreenFrameDisplayId(): number | null {
```
yagudaev: **[Codex]** **Minor (style)** — `getCurrentScreenFrameDisplayId` is exported below the `Helpers` divider and after the private helper functions `applyShow` / `resolveDisplay`, which violates the project rule that public interface lives at the top and helpers at the bottom. Move this export up alongside the other exported screen-frame functions, leaving only private helpers under the `Helpers` marker.

yagudaev (commented):
Multi-reviewer pass — partial completion.

**Codex** (gpt-5.5): 4 findings — 2 Major (IPC `event.sender` validation), 2 Minor (event listener cleanup, helper-vs-export ordering). All four posted as inline comments above.

**Gemini** (2.5-pro): hit free-tier rate limit on retry loop; produced no usable output. Will rerun later when quota resets.

Aggregated by Claude. Codex's findings were spot-checked against the source — all four are real, with line numbers adjusted slightly to land on the actual offending line in this rebased branch.

desktop/src/main/screen-frame.ts
```diff
@@ -0,0 +1,170 @@
...
+  } else {
+    frameWindow.loadFile(join(__dirname, '../renderer/index.html'), {
+      search: 'view=screen-frame',
+    })
+  }
+
+  return frameWindow
+}
+
+export function showScreenFrame(displayId?: number): void {
```
yagudaev, line 98: **[Gemini]** Major: `showScreenFrame` returns early without action if `frameWindow` is null/destroyed. While unlikely in normal flow, an OS-level event or renderer crash can destroy the window and the user would silently lose the live-share indicator until app restart. Consider lazily re-creating the frame window here when `!frameWindow || frameWindow.isDestroyed()` so the safety signal can recover. (Counter-argument: if the renderer crashed once, recreating may just crash again — at minimum log loudly so this case is observable.)

desktop/src/renderer/src/pages/ChatPage.tsx
```diff
@@ -279,8 +319,15 @@ export function ChatPage() {
...
     setIsScreenSharing(false)
     setScreenSourceName('')
+    window.electronAPI?.screenShare
+      ?.setActive?.({ active: false })
+      .catch(() => {})
   }, [])
 
+  // Keep the forward-ref in sync so track.onended (registered at
+  // capture.start time) always invokes the latest stop closure.
+  stopScreenShareRef.current = stopScreenShare
```
yagudaev: **[Gemini]** Major: `stopScreenShareRef.current = stopScreenShare` runs in the component body during render, which is a side-effect-in-render anti-pattern (problematic under StrictMode double-invoke and concurrent rendering). Move the assignment into a `useEffect` (or `useLayoutEffect` if it must be flushed before paint) so the ref is only updated after the render commits.

desktop/src/main/screen-frame.ts
yagudaev: **[Codex + Gemini]** Minor: `getCurrentScreenFrameDisplayId` (and the `currentDisplayId` it reads) appears to be dead code — it is never imported or called anywhere in the diff, never wired through IPC, and never read by another module. It also lives at the bottom of the file under the Helpers marker, which violates the project layout rule (public interface at the top, helpers at the bottom). Either remove it (and `currentDisplayId` if no other consumer needs it) or hoist it above the `// Helpers` divider with the other exports.

desktop/src/renderer/src/pages/SettingsPage.tsx
```diff
@@ -176,6 +181,11 @@ export function SettingsPage() {
     setSetting('call_bar_enabled', v ? 'true' : 'false')
   }, [])
 
+  const toggleScreenFrame = useCallback((v: boolean) => {
+    setScreenFrameEnabled(v)
```
yagudaev, line 214: **[Gemini]** Minor: `toggleScreenFrame` writes the setting to the DB and updates local state but does not notify main. If a user toggles the indicator OFF mid-share the perimeter frame stays visible until the share is stopped and restarted (and vice versa for ON). Either send an IPC signal here so main can call `hideScreenFrame()` / `showScreenFrame()` immediately, or have main re-read the setting and react when state changes.

desktop/src/renderer/src/screen-frame/screen-frame.css
```diff
@@ -0,0 +1,83 @@
...
+
+.screen-frame__border {
+  position: absolute;
+  inset: 0;
+  border: 3px solid var(--brand-accent);
+  border-radius: 12px;
+  box-sizing: border-box;
+  /* Slow, calm pulse — never alarmist. 3 s breathe between full opacity
+   * and 55 %. The animation runs on the border element only so the
+   * pill at top stays steady and readable. */
```
yagudaev: **[Gemini]** Nit: Comment still references the pill that the latest commit dropped (`fix(desktop): screen-frame rounded corners + drop pill`). Same stale wording also appears at line 18 (`border + pill`). Update both comments so the file documents the actual UI (border-only, breathing pulse) instead of the v0 design that included a pill label.

yagudaev (commented):
Multi-reviewer pass complete.

**Codex** raised 4 findings (2 Major / 2 Minor / 0 Nit).
**Gemini** raised 6 findings (2 Major / 2 Minor / 2 Nit).
**Overlap:** 1 issue raised by both (`getCurrentScreenFrameDisplayId` is dead code AND lives below the helpers divider — Codex flagged the layout, Gemini flagged dead code; merged into one inline comment).

Summary of unique findings (8 inline comments posted):

Major (4):
- IPC `screen-frame:ready` accepts any renderer (Codex)
- IPC `screen-share:setActive` accepts any renderer + unvalidated payload (Codex)
- `showScreenFrame` silently fails if window is destroyed mid-session (Gemini)
- `stopScreenShareRef.current = stopScreenShare` runs in render body (Gemini)

Minor (3):
- `track.addEventListener('ended', ...)` not removed in `ScreenCapture.stop()` (Codex)
- `getCurrentScreenFrameDisplayId` is dead + in wrong section (Codex + Gemini)
- `toggleScreenFrame` doesn't notify main, so toggle is inert mid-share (Gemini)

Nit (1):
- `screen-frame.css` comments still mention the dropped \"pill at top\" (Gemini)

Notes on tooling: Gemini hit 429 \"MODEL_CAPACITY_EXHAUSTED\" twice on `gemini-3-flash-preview` and recovered after backoff — the review eventually completed. Codex initially tried to invoke its built-in `review` skill which would have shelled out to Claude CLI; re-ran with stdin piping and an explicit \"do not invoke any skill\" instruction so the review came directly from Codex.

See inline comments for details. Aggregated by Claude.

yagudaev:
**Codex review — NEEDS FIXES before merge** (Gemini 429'd, skipped)

### Critical
- `desktop/src/main/screen-frame.ts:76` — `currentDisplayId = null` assigned in the `closed` handler but the variable is never declared in the file. Will fail TS compile / throw at runtime on window teardown. PR description claims `yarn typecheck` is clean — please re-verify.
- `desktop/src/main/screen-frame.ts:66` and `:308` — `setVisibleOnAllWorkspaces` omits `skipTransformProcessType: true`. The existing `desktop/src/main/call-bar.ts:83-89` documents this as load-bearing on macOS: without it, the app flips to `UIElementApplication` and disappears from Cmd+Tab/Dock. The frame is created at startup, so the regression is not gated to active sharing.
- `desktop/src/main/screen-frame.ts:173` — Window-source captures (`display_id` empty) fall back to `getPrimaryDisplay()`, so sharing a window on a secondary monitor draws the perimeter privacy frame around the wrong screen — a misleading "you are sharing this" signal. Either suppress the frame for window sources or resolve the actual host display.
- `desktop/src/renderer/src/screen-frame/ScreenFrame.tsx:9` — Source comment narrates abandoned implementation history ("A pill label was tried at top..."). Move to PR description / docs; project rules treat this as a blocker.

### Should fix
- `desktop/src/renderer/src/pages/SettingsPage.tsx:210` — Toggling `screen_frame_enabled` only writes the DB row; doesn't reconcile a currently-active frame. Disabling mid-share leaves the frame up; enabling mid-share doesn't show one until the next start event.
- `desktop/src/renderer/src/pages/ChatPage.tsx:310` — `track.onended` calls `stopScreenShare` without verifying the ended track still belongs to the active `ScreenCapture`; a stale ended event from an overlapping share can clear indicators on a newer share. Capture the `capture` instance in the closure and no-op unless it's still current.

Skipping merge — left for you to address.

PR #216, opened 2026-04-25
fix(desktop): keep VoiceClaw in Cmd+Tab after closing main window

yagudaev:
## Why

Closing the VoiceClaw main window made the app disappear from Cmd+Tab and the Dock. `hideMainWindow()` was calling `app.dock?.hide()`, which on macOS switches the activation policy to `accessory` and removes the app from the application switcher. The `wasOpenedAsHidden` launch-at-login path did the same thing on every login. Users reasonably expect to bring VoiceClaw back the same way they bring back any other Mac app — through Cmd+Tab — so the menu-bar-only behavior read as a regression rather than a feature.

This PR keeps the dock icon visible at all times. The tray remains as a parallel entry point. Same pattern Slack, Linear, etc. use for "open at login, hidden": suppress the window, keep the dock icon.

## Screenshots

No visual diff to capture; the change is an OS-level activation-policy effect (Cmd+Tab membership). The window itself is unchanged.

- [x] `yarn typecheck` clean
- [x] `yarn build` succeeds
- [ ] Manual smoke: launch app → close main window with red traffic light → confirm VoiceClaw still appears in Cmd+Tab and the Dock → click it (or use tray "Open VoiceClaw") → window reappears
- [ ] Manual smoke (login-at-startup): with launch-at-login enabled and `openAsHidden: true`, confirm app launches silently (no focus steal) but is reachable via Cmd+Tab and the Dock from the moment it launches
- [ ] Manual smoke: Cmd+Q still actually quits

Live UI verification was blocked by another in-progress dev Electron holding the singleton lock, so the manual checks above are still pending. The change is mechanically a removal of two `app.dock?.hide()` calls in the main process.

</details>

- `desktop/src/main/window-lifecycle.ts::hideMainWindow()` no longer calls `app.dock?.hide()`. `showMainWindow()`'s `app.dock?.show()` is left intact (cheap no-op when the dock is already visible, useful safety net for any future code path that does hide it).
- `desktop/src/main/index.ts` — `wasOpenedAsHidden` path no longer hides the dock either. Without that change, the regression would reproduce on every reboot until the user clicked the tray once.
- This menu-bar-only behavior originated in #189 (Phase 0 distribution scaffolding), not the recent floating-call-bar PR (#214). The user-visible regression became salient after #214 because the call bar normalized "the main window is closed" as a steady state.
- Reviewed by Codex + Gemini before promotion; review comments were what surfaced the `wasOpenedAsHidden` path and a stale comment in `window-all-closed`.

</details>


yagudaev:
## Code review (Codex + Gemini)

Both reviewers ran against the original `hideMainWindow()`-only diff.

### Codex
**Critical:** _none_
**Should fix:** _none_
**Nits:** stale `window-all-closed` comment claiming the dock is hidden (`desktop/src/main/index.ts:189`).

### Gemini
**Critical:** _none_
**Should fix:**
- Stale comment in `window-all-closed` (same as Codex's nit), `desktop/src/main/index.ts:186`.
- `wasOpenedAsHidden` block at `desktop/src/main/index.ts:92` still calls `app.dock?.hide()`, which contradicts the PR's intent — on every login-at-startup the regression would reproduce until the user clicked the tray.

**Nits:** _none_

### Resolution
Addressed both findings in the follow-up commit (7b5d059): dropped `app.dock?.hide()` from the `wasOpenedAsHidden` branch and refreshed the stale comments. No critical issues.

yagudaev:
## Follow-up: actual root cause was different

After the first two commits the bug persisted — VoiceClaw still missing from Cmd+Tab. `lsappinfo info <pid>` on the running Electron reported `type="UIElement"`, meaning macOS still thought the app was an accessory, even though the Info.plist has no `LSUIElement` and we no longer call `app.dock?.hide()`.

Consulted Codex (Gemini was rate-limited mid-investigation). Codex traced it to Electron's macOS source: `BrowserWindow.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true })` calls `DockHide()` internally when `visibleOnFullScreen` is `true` and `skipTransformProcessType` is not set. The call bar in `desktop/src/main/call-bar.ts` was calling exactly that at startup, which silently re-introduced the same accessory-app behavior the first commit was supposed to fix.

### Fix (commit 7c296f5)

```ts
// before
callBar.setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true })

// after
callBar.setVisibleOnAllWorkspaces(true, { skipTransformProcessType: true })
```

### Trade-off

Dropping `visibleOnFullScreen: true` means the floating call bar will no longer ride along into other apps' native-fullscreen Spaces. Staying in Cmd+Tab is the more valuable property; we can revisit cross-fullscreen visibility separately if it matters in practice.

### Reference

Electron typings (`node_modules/electron/electron.d.ts`):

> Calling setVisibleOnAllWorkspaces will by default transform the process type between UIElementApplication and ForegroundApplication … if your window is already of type UIElementApplication, you can bypass this transformation by passing true to skipTransformProcessType.

yagudaev:
## Gemini consult finally completed (it was rate-limited mid-run)

For the record — Gemini eventually pushed through the 429s and produced a different diagnosis from Codex. Two issues with it:

1. It claims the dev Electron bundle "frequently" has `LSUIElement=true`. I checked directly with `plutil -p desktop/node_modules/electron/dist/Electron.app/Contents/Info.plist` — neither `LSUIElement` nor `LSBackgroundOnly` is set. The dev bundle is a normal foreground app.
2. Its proposed fix (route `ready-to-show` through `showMainWindow()` so `app.dock?.show()` runs at startup) wouldn't have fixed the bug. Even with that change, the subsequent `setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true })` in `call-bar.ts` re-flips the activation policy to UIElement immediately afterward — that's what `lsappinfo info <pid>` reported on the running Electron.

Sticking with Codex's source-cited diagnosis and the shipped fix in `7c296f5`.

PR #221, opened 2026-04-25
fix(NAN-642): de-dupe streaming words in desktop conversation renderer

yagudaev:
## Summary
- Fixes [NAN-642](https://linear.app/nano3labs/issue/NAN-642): desktop conversation renderer duplicates every streaming word/phrase (`here. here.`, `Just Just`, `waiting waiting`, `what what`, `thinking thinking`).

## Root cause
`ChatPage.onTranscriptDelta` passed an **impure** updater to `setStreamingRole`. The updater performed a side effect — it called `setStreamingText(...)` from inside itself. React 19 `<StrictMode>` (enabled in `desktop/src/renderer/src/main.tsx`) intentionally double-invokes state updaters in development to surface impurity, so the inner `setStreamingText((prev) => prev + text)` was queued twice per delta. Net effect: every transcript delta was appended to the streaming bubble twice → doubled words.

```ts
// before — impure: side-effect inside updater, doubled by StrictMode in dev
setStreamingRole((prevRole) => {
  if (prevRole !== role) setStreamingText(text)
  else                   setStreamingText((prev) => prev + text)  // queued twice
  return role
})
```

This is dev-build-only behavior (StrictMode is a no-op in production), but `yarn dev` is the typical desktop loop, hence the report.

## Fix
Mirror `streamingRole` in a ref (`streamingRoleRef`) so the "reset vs append" decision happens synchronously in the event callback, outside any setState updater. All `setStreamingText` / `setStreamingRole` calls are now pure (constant or `prev -> next`), so StrictMode's double-invocation has no observable effect. Other `setStreamingRole` call sites (`onTurnStarted`, `onToolProgress`) keep the ref in sync.

## Files changed
- `desktop/src/renderer/src/pages/ChatPage.tsx`

## Relation to NAN-637
**Separate bug.** NAN-637 is about timestamp/segment metadata (e.g. `[1m19s70ms]`) appearing inline in transcript text — that's an upstream relay/adapter content issue. NAN-642 is purely a renderer-side React StrictMode/impure-updater issue and is unrelated.

## Test plan
- [x] `yarn typecheck` (desktop) — clean
- [x] `yarn build` (desktop) — clean
- [ ] Live test: `yarn dev` desktop + dev relay, run a real Gemini call, confirm no doubled words during or after streaming.
- [ ] Verify `transcript.done` final commit still renders once (no race with the streaming bubble) — manual.

I have not exercised a live conversation against a relay from this worktree; the fix is verified via static analysis + StrictMode semantics + typecheck/build. Recommend a quick smoke before merge.

### DIFF
--- a/desktop/src/renderer/src/pages/ChatPage.tsx
+++ b/desktop/src/renderer/src/pages/ChatPage.tsx
@@ -171,12 +171,14 @@ export function ChatPage() {
       setIsCallActive(false)
       setIsThinking(false)
       setStreamingText('')
+      streamingRoleRef.current = 'assistant'
     },
     onDisconnect: () => {
       setIsCallActive(false)
       setIsConnecting(false)
       setIsThinking(false)
       setStreamingText('')
+      streamingRoleRef.current = 'assistant'
     },
     onError: (message) => {
       console.error('[ChatPage] Relay error:', message)
--- a/desktop/src/renderer/src/pages/ChatPage.tsx
+++ b/desktop/src/renderer/src/pages/ChatPage.tsx
@@ -35,6 +35,11 @@ export function ChatPage() {
   const [isThinking, setIsThinking] = useState(false)
   const [streamingText, setStreamingText] = useState('')
   const [streamingRole, setStreamingRole] = useState<'user' | 'assistant'>('assistant')
+  // Mirror of streamingRole readable synchronously in event callbacks. Used by
+  // onTranscriptDelta so we can decide "reset vs append" without putting an
+  // impure side effect inside a setState updater (which StrictMode would
+  // double-invoke and cause word duplication).
+  const streamingRoleRef = useRef<'user' | 'assistant'>('assistant')
   const [showLatency, setShowLatency] = useState(false)
   const [connectionError, setConnectionError] = useState('')
   const [activeRealtimeModel, setActiveRealtimeModel] = useState('')
@@ -106,15 +111,18 @@ export function ChatPage() {
       setIsCallActive(true)
     },
     onTranscriptDelta: (text, role) => {
-      setStreamingRole((prevRole) => {
-        if (prevRole !== role) {
-          // Role switched (user→assistant or vice versa) — reset the buffer
-          setStreamingText(text)
-        } else {
-          setStreamingText((prev) => prev + text)
-        }
-        return role
-      })
+      // Role/buffer reconciliation must NOT happen inside a setState updater —
+      // React StrictMode runs updaters twice (in dev) to surface impure ones,
+      // which double-fires the side-effect setStreamingText(...) and produced
+      // the doubled-words bug (NAN-642). The current role lives in a ref so
+      // we can read it synchronously here without conjuring an updater.
+      if (streamingRoleRef.current !== role) {
+        streamingRoleRef.current = role
+        setStreamingRole(role)
+        setStreamingText(text)
+      } else {
+        setStreamingText((prev) => prev + text)
+      }
       if (role === 'assistant') setIsThinking(false)
     },
     onTranscriptDone: async (text, role) => {
@@ -135,6 +143,7 @@ export function ChatPage() {
     onTurnStarted: () => {
       setIsThinking(false)
       setStreamingText('')
+      streamingRoleRef.current = 'user'
       setStreamingRole('user')
     },
     onTurnEnded: () => {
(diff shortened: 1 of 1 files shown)


yagudaev (commented):
## Codex review (NAN-642)

**Verdict: COMMENT** (no blockers; ship after considering the NITs below.)

### Root cause analysis: confirmed correct
React 19 StrictMode is enabled (`desktop/src/renderer/src/main.tsx:23`). The original `setStreamingRole((prevRole) => { ...; setStreamingText(...); return role })` packed a side effect inside an updater. StrictMode invokes updaters twice in dev to surface impurity, so the inner `setStreamingText(...)` got enqueued twice per delta — exactly the "here. here." / "Just Just" symptom.

### Fix assessment: correct, no race
- The ref is read/written synchronously inside a WebSocket message callback (outside React's render lifecycle), so there's no double-invocation.
- All `setState` calls are now pure (single state, value or pure functional updater) — StrictMode-safe.
- Role-mismatch reset still fires correctly on the first delta after `onTurnStarted` (which initializes the ref to `'user'`) and after `onToolProgress` (which sets it to `'assistant'`).
- No stale-ref hazard: refs always reflect the latest write at event time.

### NITs (non-blocking)

**NIT 1 — `desktop/src/renderer/src/pages/ChatPage.tsx:170-180`**
`onSessionEnded` and `onDisconnect` clear `streamingText` but don't reset `streamingRoleRef.current` (or `streamingRole` state — pre-existing). On a new session, if the first delta arrives with the same role as the last cleanup left behind, we hit the append branch on an already-empty `streamingText`, which is harmless today but fragile. Suggest:
```ts
streamingRoleRef.current = 'assistant'
setStreamingRole('assistant')
```
in both handlers.

**NIT 2 — pattern duplication**
`onTurnStarted` (line 146-147) and `onToolProgress` (line 166-167) each have to remember to update both `streamingRoleRef.current` and `setStreamingRole`. A tiny local helper at the bottom of the file (`setRole(role: 'user' | 'assistant')` that updates ref + state) would eliminate the chance of future drift. Helpers-at-bottom convention already in use here.

### QUESTION
Was the fix verified on a dev build with StrictMode enabled (i.e. could you reproduce "here. here." on `main` and confirm it's gone on this branch)? If so, mention that in the PR description so future debuggers don't have to reverse-engineer the repro.

### Style: compliant
No semicolons, helpers at bottom under the `// Helpers` divider, public component at top. Good.

### Alternative considered
A `useEffect` reacting to `streamingRole` changes was considered but would add a render cycle of latency to text streaming — worse than the ref. The ref pattern is the right call.

_(Codex CLI was used as the reviewer.)_

yagudaev:
## Sonnet cross-model review: LGTM (would APPROVE)

> Note: GitHub prevents self-approval so this is posted as a comment, but the verdict is **APPROVE**.

---

### Root-cause analysis: Correct

StrictMode is confirmed enabled in `desktop/src/renderer/src/main.tsx` (lines 23 and 38) for both the main app and the call-bar window. The old code called `setStreamingText(...)` as a side-effect inside a `setStreamingRole` updater — a textbook impure updater that StrictMode's double-invocation exposes. The duplication was real and dev-only, as claimed.

### Does the ref fix work? Yes

`use-realtime.ts` already uses the `callbacksRef.current = callbacks` pattern (line 89) so there is no stale-closure risk on the callback side. The ref is read synchronously in the JS event handler — single-threaded, no race possible.

### Production impact

StrictMode is dev-only. The old code was correct in production; the fix is a no-op in production. Zero prod regression risk.

### Simpler alternatives considered

Two alternatives exist but are not clearly better:
1. **Local variable outside updater (no ref)** — would work too, but requires the closure to re-capture `streamingRole` state on every render. The ref approach avoids that churn and is idiomatic for "synchronously readable copy of state."
2. **Collapse role + text into a single reducer** — cleaner long-term but a larger refactor outside the scope of a targeted bug fix. Not worth it here.

The ref mirror is the right call.

### NITs (non-blocking)

**`ChatPage.tsx` lines 170–180** — `onSessionEnded` and `onDisconnect` clear `streamingText` but leave `streamingRoleRef.current` stale. Harmless in practice (next `onTurnStarted` resets it; first `onTranscriptDelta` after reconnect self-corrects via the role-mismatch branch), but inconsistent with the write-both-together discipline established in `onTurnStarted` (line 146–147) and `onToolProgress` (line 166–167). A one-liner `streamingRoleRef.current = 'assistant'` in each handler would close the gap.

**Test plan** — Author explicitly flags the live device test is unchecked. Static analysis is convincing here, but a quick smoke before merge is recommended per the PR's own checklist.

No blockers. Fix is correct, well-motivated, and well-commented.

PR #222, opened 2026-04-25
fix(NAN-641): make brain gateway URL primary, hide legacy Vapi setup

yagudaev:
## Summary

Closes [NAN-641](https://linear.app/nano3labs/issue/NAN-641).

The mobile app still showed "Please configure your Vapi Public Key and Assistant ID in Settings first" on the chat screen and exposed a "Voice Pipeline" / "Relay Server URL" card that did not point at the brain gateway. Vapi/Custom-pipeline are dead code paths after a previous cleanup but the leftover copy was misleading new users.

This PR:

- Settings: rename the card "Voice Pipeline" -> "Brain Gateway", and the URL field "Relay Server URL" -> "Brain Gateway URL". Rewrite the Setup blurb so step 1 says "Run your brain gateway (relay-server)..." and steps 2-3 reference the gateway URL + matching API key. Replaces the leftover "Use Custom Pipeline or Vapi mode" line in the latency-stats empty state.
- Chat: drop the now-dead `voice_mode === 'custom' | 'vapi'` branches in `toggleCall` so we always start a realtime call. Replace the two "Please configure your Vapi Public Key and Assistant ID..." messages in `startVapiCall` (now unreachable but still in source) with copy that points at Brain Gateway URL setup. Update the text-chat empty-config error to mention "Brain Gateway".

Vapi / custom pipeline code paths are kept (still unreachable through UI) so this is a low-risk hide-not-delete change. A follow-up can rip out the `ExpoVapiModule` / `ExpoCustomPipelineModule` imports + helpers entirely once we are confident no one is still on those paths.

## Acceptance criteria

- [x] settings clearly point to brain gateway URL setup
- [x] Vapi legacy path hidden (no UI surface, dead code paths)
- [x] custom pipelines hidden (no UI surface)
- [x] chat screen no longer blocks on Vapi public key / assistant id messaging

## Test plan

- [x] `yarn tsc --noEmit` in `mobile/` passes
- [ ] `APP_VARIANT=development npx expo run:ios --device` on the test iPhone
- [ ] Open Settings: "Brain Gateway" card visible, URL field labeled "Brain Gateway URL", setup hint mentions brain gateway
- [ ] Open Chat: text input with no brain config shows the new "Text chat needs a brain gateway URL..." copy
- [ ] Tap call button with realtime relay configured: starts a realtime call (Vapi/custom paths unreachable)

## Notes for the human reviewer

- I was not able to flash the iPhone test device from this CLI session, so the only verification I ran was `yarn tsc --noEmit`. Please do a quick smoke-run on device before merging.
- The mobile `lib/chat.ts` text-chat path still uses `brain_api_url` (a separate KV from the realtime relay URL), and there is no UI to set `brain_api_url`. The new error copy directs the user to the Brain Gateway settings; if you want text chat fully usable the next step is to either (a) add a second field for `brain_api_url`, or (b) hide text-chat send entirely until brain gateway is wired through to text mode. I left this out of scope for this ticket.

### DIFF
--- a/mobile/app/(tabs)/index.tsx
+++ b/mobile/app/(tabs)/index.tsx
@@ -5,14 +5,14 @@ import { Icon } from '@/components/ui/icon'
 import { Input } from '@/components/ui/input'
 import { Text } from '@/components/ui/text'
 import { addMessage, createConversation, getLatestConversation, getMessages, getSetting, type Message, type LatencyData, type ProviderInfo } from '@/db'
-import { getApiConfig, streamCompletion } from '@/lib/chat'
+import { getApiConfig, streamRealtimeText } from '@/lib/chat'
 import { BRAND } from '@/lib/brand'
 import { compactMessages } from '@/lib/compact'
 import { useConversationContext } from '@/lib/conversation-context'
 import { maybeGenerateTitle } from '@/lib/title'
 import { useCallSounds } from '@/lib/sounds'
 import { usePipeline } from '@/lib/use-pipeline'
-import { useRealtime, type RmsMetrics } from '@/lib/use-realtime'
+import { useRealtime, getProviderForRealtimeModel, type RmsMetrics } from '@/lib/use-realtime'
 import { useTranscriptBuffer } from '@/lib/use-transcript-buffer'
 import { useAutoReconnect } from '@/lib/use-auto-reconnect'
 import ExpoCustomPipelineModule from '@/modules/expo-custom-pipeline/src/ExpoCustomPipelineModule'
@@ -644,19 +644,38 @@ export default function ChatScreen() {
       return
     }
 
-    const { apiKey, model, apiUrl } = await getApiConfig()
-    if (!apiKey || !apiUrl) {
-      await addMessage(conversationId, 'assistant', 'Text chat over the Brain Gateway is not wired up yet — tap the mic to start a voice call.')
+    const serverUrl = (await getSetting('realtime_server_url')) || 'ws://localhost:8080/ws'
+    const apiKey = await getSetting('realtime_api_key')
+    if (!apiKey) {
+      await addMessage(conversationId, 'assistant', 'Configure Brain Gateway URL in Settings to start chatting.')
       await loadMessages()
       return
     }
 
-    setIsThinking(true)
-    const allDbMessages = await getMessages(conversationId)
-    const allMessages = await compactMessages(conversationId, allDbMessages, apiKey, model, apiUrl)
-
+    const model = normalizeRealtimeModel(await getSetting('realtime_model'))
+    const voice = normalizeRealtimeVoice(model, await getSetting('realtime_voice'))
+    const provider = getProviderForRealtimeModel(model)
     const systemPrompt = (await getSetting('system_prompt')) || ''
-    streamCompletion(allMessages, apiKey, model, apiUrl, systemPrompt, conversationId, {
+    const timezone = Intl.DateTimeFormat().resolvedOptions().timeZone
+    const locale = Intl.DateTimeFormat().resolvedOptions().locale
+
+    const recentMessages = (await getMessages(conversationId))
+      .filter((m) => m.role === 'user' || m.role === 'assistant')
+      .slice(-20, -1)
+      .map((m) => ({ role: m.role as 'user' | 'assistant', text: m.content }))
+
+    setIsThinking(true)
+    streamRealtimeText(userInput, {
+      serverUrl,
+      apiKey,
+      model,
+      voice,
+      provider,
+      sessionKey: `voiceclaw:${conversationId}`,
+      instructionsOverride: systemPrompt || undefined,
+      conversationHistory: recentMessages.length > 0 ? recentMessages : undefined,
+      deviceContext: { timezone, locale },
+    }, {
       onToken: (text) => {
         setIsThinking(false)
         setStreamingRole('assistant')
@@ -664,6 +683,7 @@ export default function ChatScreen() {
       },
       onDone: async (text) => {
         setStreamingText(null)
+        setIsThinking(false)
         await addMessage(conversationId, 'assistant', text || 'No response received.')
         await loadMessages()
         maybeGenerateTitle(conversationId)
--- a/mobile/app/(tabs)/index.tsx
+++ b/mobile/app/(tabs)/index.tsx
@@ -417,20 +417,6 @@ export default function ChatScreen() {
   cancelReconnectRef.current = cancelReconnect
   triggerReconnectRef.current = triggerReconnect
 
-  const ensureVapiReady = useCallback(async (): Promise<boolean> => {
-    if (vapiReady) return true
-    const apiKey = (await getSetting('vapi_public_key')) || (await getSetting('vapi_api_key'))
-    if (!apiKey) return false
-    try {
-      await ExpoVapiModule.initialize(apiKey)
-      setVapiReady(true)
-      return true
-    } catch (e) {
-      console.warn('Vapi init failed:', e)
-      return false
-    }
-  }, [vapiReady])
-
   useEffect(() => {
     const subs = [
       ExpoVapiModule.addListener('onCallStart', () => {
@@ -660,7 +646,7 @@ export default function ChatScreen() {
 
     const { apiKey, model, apiUrl } = await getApiConfig()
     if (!apiKey || !apiUrl) {
-      await addMessage(conversationId, 'assistant', 'Text chat needs a brain gateway URL. Open Settings and complete the Brain Gateway setup, or start a voice call instead.')
+      await addMessage(conversationId, 'assistant', 'Text chat over the Brain Gateway is not wired up yet — tap the mic to start a voice call.')
       await loadMessages()
       return
     }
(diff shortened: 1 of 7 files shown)


mobile/app/(tabs)/index.tsx
```diff
@@ -660,7 +660,7 @@ export default function ChatScreen() {
 
     const { apiKey, model, apiUrl } = await getApiConfig()
     if (!apiKey || !apiUrl) {
-      await addMessage(conversationId, 'assistant', 'Please configure your brain agent API URL and key in Settings first.')
+      await addMessage(conversationId, 'assistant', 'Text chat needs a brain gateway URL. Open Settings and complete the Brain Gateway setup, or start a voice call instead.')
```
yagudaev: **BLOCKER — empty-state copy creates a setup loop**

This new copy says "complete the Brain Gateway setup," but the only visible Brain Gateway card in Settings (mobile/app/(tabs)/settings.tsx:483-510) writes to `realtime_server_url` / `realtime_api_key`, not `brain_api_url` / `brain_api_key`. `getApiConfig()` (mobile/lib/chat.ts:151-157) reads only the `brain_api_*` keys, and `updateBrainApiUrl` / `updateBrainApiKey` exist in `settings.tsx` but are not rendered into any JSX, so a brand-new user has no way to populate them.

So the new flow is: open empty chat → tap Send text → see this error → go to Settings → fill out the only Brain Gateway card visible → come back → tap Send → hit this same error again. That is exactly the urgency NAN-641 is fixing for the call path; we should not ship a regression of the same shape on the text path.

Two acceptable resolutions:
1. Hide / disable the text input + send button when `brain_api_url` is unset (and update the empty-state CTA to “Start a voice call”). This matches the worker's "mobile is realtime-only" framing.
2. Or have the visible Brain Gateway card also write `brain_api_url` / `brain_api_key` in addition to the realtime keys, so one setup actually wires both paths.

Option 1 is the smaller and cleaner change for this PR.

yagudaev:
## Codex review — REQUEST CHANGES (self-PR, posted as comment)

Reviewer: codex-cli 0.124.0. One inline BLOCKER posted on `mobile/app/(tabs)/index.tsx:663`. GitHub disallows request-changes on your own PR, so the formal review is filed as a `--comment` and the verdict is captured here.

### Verdict on the worker's two open questions

**(1) Text-chat `brain_api_url` gap — BLOCKER, do not defer.** The new empty-state copy ("complete the Brain Gateway setup") points at a card that doesn't actually populate the keys text-chat reads. `updateBrainApiUrl` / `updateBrainApiKey` exist in `settings.tsx` but are never rendered, so a fresh user can complete the only visible setup and still loop on the same error. Fix in this PR — easiest path is to hide / disable the text input + Send button when `brain_api_url` is missing (mobile is realtime-only per `mobile/CLAUDE.md`). See inline.

**(2) Dead Vapi/custom-pipeline code — fine to defer.** `toggleCall` at `mobile/app/(tabs)/index.tsx:705-738` forces `realtime` and never reads `voice_mode`. The only remaining `voice_mode` references are the migration-to-`realtime` shim in `settings.tsx:208-214` and the unused `updateVoiceMode` setter at `settings.tsx:282-285`. No runtime path reaches `startVapiCall` / `startCustomPipelineCall` / `ensureVapiReady` anymore. Safe to ship and clean up later.

### NITs

- `mobile/app/(tabs)/index.tsx:936` — mic-setup error still says "Realtime settings"; Settings now labels that card "Brain Gateway." Consider "...in Brain Gateway settings first."
- `mobile/app/(tabs)/index.tsx:753, 762` — `startVapiCall` still emits user-visible strings containing "Vapi". Unreachable today, but trips a strict "Vapi" grep. Cleanup with the dead-code follow-up.
- `mobile/app/(tabs)/settings.tsx:282-285` — `updateVoiceMode` is unreachable (no JSX renders it). Tiny dead-code item.

### Style / mechanical

- No semicolons added by this diff.
- No public-interface vs helper reordering issues in the diff.
- `git diff --check origin/main -- mobile/` is clean. No `Public Key` / `Assistant ID` strings left user-visible.

### Simpler approach

If "mobile is realtime-only" is the real intent, the cheapest combined fix to (1) and (2):
1. Hide the text input + Send button (or gate them on `brain_api_url` existing) and surface a one-line CTA pointing at the voice-call button.
2. Delete `startVapiCall` / `startCustomPipelineCall` / `ensureVapiReady` and the `voice_mode` setting entirely in a follow-up.

Also worth removing the unused `updateBrainApiUrl` / `updateBrainApiKey` setters from `settings.tsx` in this PR, since they have no UI behind them — keeping them implies a future card that doesn't exist.

yagudaev (commented):
Codex review posted as PR comment (request-changes blocked on own PR). Net verdict: REQUEST CHANGES — blocker on text-chat empty-state copy looping users back to the same error. See inline comment on `mobile/app/(tabs)/index.tsx:663` and the summary comment.

yagudaev:
## NAN-641 follow-up — STOP and report

### What changed in commit `152d34a`

Cleanups only. No routing change yet.

- Removed dead `startVapiCall`, `startCustomPipelineCall`, `ensureVapiReady` from `mobile/app/(tabs)/index.tsx` (defined, never called from any reachable path).
- Removed dead `updateVoiceMode` setter from `mobile/app/(tabs)/settings.tsx`.
- Mic-setup error: "Realtime settings" → "Brain Gateway settings".
- Text-chat empty-state copy is now honest about the missing wire-up: *"Text chat over the Brain Gateway is not wired up yet — tap the mic to start a voice call."* (Previously implied that completing the visible setup would unblock text chat, which it does not.)

### Why I did NOT route text through realtime in this PR

The original direction — *"the realtime models accept text input too, just wire text-chat through the same path"* — is correct at the upstream-provider level (Gemini's `realtimeInput.text`, OpenAI's `conversation.item.create + input_text`). Both adapters in `relay-server/src/adapters/` already use this internally for `injectContext`.

**But the relay's client-facing protocol does not expose it.** `relay-server/src/types.ts` has no `TextInputEvent` in the `ClientEvent` union, and `session.ts handleMessage` has no case to forward text → `adapter.injectContext`. Adding one is small (~10 lines: one type, one switch case, one mobile send call) but it crosses the relay-server boundary.

The original prompt was explicit: *"If it requires non-trivial relay-server work, STOP and report — that's a follow-up ticket, not this PR."* I'm choosing to honor that gate even though the change is small, because expanding the wire-protocol surface deserves its own ticket where the design (event name, response semantics, error handling, whether to hold the WS open or close after one turn, how it interacts with an active voice session) gets reviewed on its own merits.

### Does the realtime adapter already support text input upstream?
- Gemini: yes — `injectContext` in `relay-server/src/adapters/gemini.ts:351` sends `realtimeInput.text` and Gemini will generate a response.
- OpenAI: yes — `relay-server/src/adapters/openai.ts:91` sends `conversation.item.create` with `{type:"input_text", text}` then calls `requestResponse`.
- xAI: inherits from OpenAI adapter, same path.

So the upstream side is ready. **Only the relay client-facing protocol needs the new event.** I will open a follow-up ticket if you confirm the direction.

### Orphan keys

Not removed in this commit. `mobile/lib/chat.ts:152-156` still reads `brain_api_url` / `brain_api_key` / `openclaw_*` fallbacks — but with the new copy, that path is unreachable through any UI a fresh user can configure. Pulling the orphan reads + the entire `streamCompletion` SSE path is best done in the same PR that adds the realtime-text routing, so the diff tells one coherent story.

### Verification
- `cd mobile && yarn tsc --noEmit` — passes.
- Did not flash to device from this CLI session.

### What I'd like a human decision on before merge
1. Ship this cleanup as-is (NITs + honest text-chat copy) and open a follow-up to route text through realtime?
2. Or expand scope here: add `text.input` to relay protocol + a `sendText` on `useRealtime` and handle it in `sendMessage`?

I lean toward (1) — the relay-protocol change deserves its own review.

yagudaev:
Round 3: wired mobile text-chat through the realtime path.

**Commit:** `8ff57fa`

**Relay protocol:**
Added `TextInputEvent` to `ClientEvent`:
```ts
export interface TextInputEvent {
  type: "text.input"
  text: string
}
```
`session.handleMessage` forwards it to `this.adapter?.injectContext(event.text)`. Both adapters' `injectContext` already trigger a model response (Gemini via `realtimeInput.text`, OpenAI via `conversation.item.create` + `response.create`), so `transcript.delta` / `transcript.done` flow back through the existing pipeline with no separate "go" signal.

**Mobile:**
- `lib/chat.ts` — new `streamRealtimeText(text, opts, cbs)` helper. Opens a one-shot WebSocket to `realtime_server_url`, sends `session.config` + `text.input`, accumulates assistant `transcript.delta`s, resolves on `transcript.done` (or `turn.ended` fallback). We never call `playAudio` in this path so any `audio.delta` is silently dropped. Removed `brain_api_*` reads from `getApiConfig`.
- `lib/use-realtime.ts` — exported `getProviderForRealtimeModel` so the helper can pick the right provider from `realtime_model`.
- `app/(tabs)/index.tsx` `sendMessage` — replaced the dead SSE path with `streamRealtimeText`. Passes the last 19 messages as `conversationHistory` and the system prompt as `instructionsOverride`. Inline error now reads "Configure Brain Gateway URL in Settings to start chatting" — true after this change.
- `app/(tabs)/settings.tsx` — dropped the orphan `brainApiKey/brainApiUrl` state and `updateBrainApi*` setters (they had no UI bindings and only wrote keys nothing read).

**Tested:**
- `relay-server`: `npx tsc --noEmit` — clean
- `mobile`: `npx tsc --noEmit` — clean

**Not tested:**
- End-to-end smoke (relay running + device sending a text message and seeing a streamed response). No relay or simulator was running in this worktree, so this is a code-only change waiting on a manual smoke. The wire flow is straightforward (`session.config` → `session.ready` → `text.input` → `transcript.delta` → `transcript.done`); the parts that warrant verification are (a) Gemini's `realtimeInput.text` triggering a response in a session that has only ever received text input (no prior audio), and (b) the user's typed message not appearing in the relay's `transcript[]` log used by `syncTranscriptToBrain` at session end — that's a known gap but doesn't break this feature.

**Surprises:**
- `compactMessages` and `maybeGenerateTitle` still use `getApiConfig`'s `apiKey/apiUrl` (now pulling from `openclaw_api_*` only, which isn't in the UI). They will silently no-op when those settings are absent — title-gen and compaction become best-effort, which matches the round-2 cleanup philosophy. Refactoring those to also use the realtime path would be a separate change.

yagudaev (commented):
Codex round-3 review — flagging one BLOCKER (Gemini context loss) plus follow-ups. GitHub won't let me REQUEST_CHANGES on my own PR, so this is filed as a comment with a do-not-merge recommendation until the blocker is resolved or explicitly accepted.

## BLOCKER

**Gemini text turns drop conversation history every turn.** `streamRealtimeText` opens a fresh WS per text turn and forwards `conversationHistory` in `session.config`, but only the OpenAI adapter consumes it (`relay-server/src/adapters/openai.ts:208`). The Gemini adapter has zero `conversationHistory` handling — `rg conversationHistory relay-server/src/adapters/gemini.ts` returns no hits. Since the default model is `gemini-3.1-flash-live-preview`, every text reply will see only the current `text.input` with no prior context ("send another one" → model has no idea what "another" refers to).

Two reasonable fixes:
- Mirror OpenAI's pattern: append the trimmed history into the system instructions in Gemini's `configureSession`-equivalent path, OR
- Send each prior `{role,text}` upstream via `realtimeInput.text` before the live user `text.input` so the model has turn-level context.

This is the only item I'd block on; it makes the new text path effectively broken for repeat turns on the default provider.

## NIT (defer)

- `mobile/lib/title.ts:14` — `maybeGenerateTitle` silently no-ops because `getApiConfig` reads orphan `openclaw_api_key` / `openclaw_api_url` settings the Settings UI no longer writes. New text chats stay titled "New Conversation"; the catch-block timestamp fallback only fires on thrown errors, not on the early-return-when-no-key path. Worth a follow-up; not a release blocker since chats remain functional.
- `mobile/lib/compact.ts` — same `getApiConfig` issue, but `compactMessages` is no longer reachable from the realtime text path, so it's dead-on-arrival noise.
- `mobile/app/(tabs)/index.tsx:641` — leftover `if (isCallActive && vapiReady) { await ExpoVapiModule.sendMessage(...) }` branch in `sendMessage`. `setVapiReady` is never called now, so it's dead but harmless. Clean up alongside the other Vapi removals.

## QUESTION

- `relay-server/src/session.ts:384` — `text.input` injects into the model but neither adapter pushes the typed user message into its `this.transcript`. `syncTranscriptToBrain()` at end-of-session therefore sends assistant-only text to the brain. Logged as a known follow-up; raising here only so it isn't lost. Easy fix: append `{role:'user', text}` to the adapter transcript right after `injectContext`, or expose an adapter hook.

## CONFIRMED CLEAN

- `text.input` typing/routing: `relay-server/src/types.ts:14,80`, handled at `session.ts:377-386`.
- `injectContext` triggers a model response on both adapters: Gemini via `realtimeInput.text` (auto-responds); OpenAI via `conversation.item.create` + `requestResponse("injectContext")`.
- Mobile `sendMessage` now uses `streamRealtimeText` (`mobile/app/(tabs)/index.tsx:668`).
- `rg "brain_api_url|brain_api_key"` → zero hits.
- No semicolons added; helpers remain at file bottom.
- `tsc --noEmit` green for both relay-server and mobile.

PR #230, opened 2026-04-25
feat(NAN-640): inject recent history plus summary on voice session resume

yagudaev:
## Summary

When a voice session restarts with `conversationHistory`, the relay now splits it into the **last 8 turns verbatim** plus a **summarized preamble** of older messages — preserving continuity without flooding the model's context window. Closes [NAN-640](https://linear.app/nano3labs/issue/NAN-640).

## What changed

- **New `relay-server/src/history.ts`** — provider-aware split + summarization.
  - Gemini sessions use `gemini-2.5-flash`'s `generateContent` REST endpoint.
  - OpenAI / xAI sessions use `gpt-4o-mini`'s `chat/completions`.
  - Falls back to the OpenAI text endpoint, then to a truncated raw transcript, so a flaky summarizer can never block the session.
- **OpenAI adapter (`openai.ts`)** — replaces the prior "stuff history into instructions with a 3000-char trim" path with one `conversation.item.create` per recent turn (role-correct content type: `input_text` for user, `text` for assistant) plus the summary as a section in the system instruction. Cleared after first injection so timer-based rotation does not replay it. xAI inherits this via `XAIAdapter`.
- **Gemini adapter (`gemini.ts`)** — now consumes `conversationHistory`. Summary is appended to `systemInstruction` at setup. The recent turns are sent as a single `clientContent { turns, turnComplete: false }` message right after `setupComplete` (matching the [official Gemini Live "Prefilling a conversation context" use case](https://github.com/googleapis/js-genai/blob/main/docs/classes/live.Session.html)). Resumption-handle reconnects do **not** replay history.
- **Docs (`relay-server.mdx`)** — new "Conversation Resume" section documents the split, the per-provider summarizer model, and the per-provider injection mechanics.

## N = 8 — rationale

8 turns (16 messages: 8 user + 8 assistant) is a doubling of the prior 3k-char cap (~5 turns). At ~100-200 tokens per voice turn, 16 messages ≈ 2-4k tokens — well under OpenAI Realtime's 128k window and negligible against Gemini Live's 1M. Larger N would not improve continuity meaningfully (older turns rarely surface in the next exchange) but would extend the per-event injection workload on every resume, measurably raising setup latency. Sources used to set the N:

- [OpenAI Realtime docs](https://platform.openai.com/docs/guides/realtime) — `conversation.item.create` is the supported channel for prefilling history; cost is per item, so latency scales with item count.
- [Gemini Live `sendClientContent`](https://github.com/googleapis/js-genai/blob/main/docs/classes/live.Session.html) — explicitly lists "prefilling a conversation context" as a use case for `clientContent` with `turnComplete: false`.

## Test plan

- [ ] Typecheck: `cd relay-server && yarn build` passes (verified locally).
- [ ] Docs build: `cd docs && yarn build` passes (verified locally).
- [ ] Smoke: start a Gemini voice session, exchange >8 turns, restart the session — confirm the model recalls early-conversation facts (delivered via summary preamble) and recent phrasing (delivered via verbatim turns).
- [ ] Smoke: same flow on OpenAI Realtime / Grok Voice.
- [ ] Failure case: revoke `OPENAI_API_KEY` while keeping `GEMINI_API_KEY` set, exercise an OpenAI session with >8 turns — relay should fall back to truncated raw transcript and continue, not crash.

### DIFF
--- a/desktop/src/renderer/src/pages/ChatPage.tsx
+++ b/desktop/src/renderer/src/pages/ChatPage.tsx
@@ -209,6 +209,14 @@ export function ChatPage() {
     const inputDeviceId = (await getSetting('input_device_id')) || undefined
     const outputDeviceId = (await getSetting('output_device_id')) || undefined
 
+    const convId = conversationIdRef.current
+    const conversationHistory = convId
+      ? (await getMessages(convId))
+          .filter((m) => m.role === 'user' || m.role === 'assistant')
+          .slice(-200)
+          .map((m) => ({ role: m.role as 'user' | 'assistant', text: m.content }))
+      : []
+
     setActiveRealtimeModel(model)
     realtime.start({
       serverUrl,
@@ -225,6 +233,7 @@ export function ChatPage() {
         locale: navigator.language,
         deviceModel: 'Desktop (Electron)',
       },
+      conversationHistory: conversationHistory.length > 0 ? conversationHistory : undefined,
       tracingEnabled,
     })
   }, [realtime])
--- a/docs/src/content/docs/relay-server.mdx
+++ b/docs/src/content/docs/relay-server.mdx
@@ -88,7 +88,21 @@ Fields:
 - `model` -- model identifier (default: `gemini-3.1-flash-live-preview` for Gemini, `grok-voice-think-fast-1.0` for xAI, `gpt-realtime-mini` for OpenAI)
 - `brainAgent` -- `"enabled"` or `"none"`
 - `apiKey` -- must match `RELAY_API_KEY` if set on the server
-- `conversationHistory` -- prior messages to inject as context (for continuing conversations)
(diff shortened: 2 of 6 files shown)


yagudaev:
Re Codex QUESTION on `openai.ts:254` (clearing resume history): yes, intentional. After the first session (with summary + recent turns injected), `this.transcript` accumulates new turns. When `ROTATION_INTERVAL_MS` fires ~50 minutes later, `summarizeTranscript()` is built from that fresh transcript — by then it represents the current 50-min window which is what the rotated session should resume from. The pre-resume context is intentionally left to age out: it was already summarized once, and summarizing a summary on every rotation would compound information loss. If we ever need the original pre-resume context to persist past the first rotation, the right place is `summarizeTranscript()` itself (concatenate the original summary), but that's out of scope for NAN-640.

PR #264, opened 2026-04-29
fix(gemini): rebuild resume preamble on forced-fresh reconnect

yagudaev:
## Why

Today's SF-trip session lost ~25 minutes of in-call context after a forced-fresh reconnect. Trace: post-resume watchdog detected a poisoned handle → forced a fresh session (no resumption handle) → new session's `systemInstruction` was the preamble built once at `connect()` from `config.conversationHistory`, so it only saw the call-start history, not anything from the live conversation.

After PR #255 (idle resume false-positive fix) the forced-fresh path is rare, but when it does fire, it's still lossy. OpenAI's `rotate()` already updates its own `resumeSummary` / `resumeRecentTurns` before reopening (PR #260) — Gemini's analog needed the same treatment.

## Fix

- Extract the preamble construction in `connect()` into a `rebuildResumePreamble(history)` helper.
- At the top of `reconnect()` when `forceFresh === true`, rebuild the preamble from `[...config.conversationHistory, ...this.transcript]` so the new Gemini session sees the original turns AND every in-call exchange accumulated since connect.
- Handle-based reconnects skip the rebuild — Gemini restores in-call state from the resumption handle, so re-folding turns into systemInstruction would just bloat the prompt.

`this.transcript` only accumulates from server events post-connect, so there's no double-count risk against `config.conversationHistory`.

## Tests

New file `relay-server/test/adapters/gemini-forced-fresh-preamble.test.ts` (3 cases):

1. Forced-fresh reconnect rebuilds the preamble from initial + in-call turns — asserts the resumed setup's `systemInstruction` text contains both the call-start history and post-connect transcript turns.
2. Handle-based reconnect leaves the preamble unchanged.
3. Forced-fresh with no in-call turns is a no-op (resumed preamble equals initial preamble).

`yarn workspace relay-server test` → 62 passed, 2 skipped (15 files).
`yarn workspace relay-server tsc --noEmit` → clean.

## Test plan

- [x] `yarn workspace relay-server test` passes
- [x] `yarn workspace relay-server tsc --noEmit` clean
- [ ] Live: trigger a forced-fresh reconnect (poisoned-handle simulation) mid-call and confirm the model continues with full context

### DIFF
--- a/relay-server/src/adapters/gemini.ts
+++ b/relay-server/src/adapters/gemini.ts
@@ -250,6 +250,16 @@ export class GeminiAdapter implements ProviderAdapter {
     this.currentlyResumable = false
     this.rotateAfterToolCalls = false
 
+    // Finalize any in-flight transcription before rotating. Gemini's resumed
+    // session replays transcription from its last checkpoint, so if we leave
+    // currentAssistantText / currentUserText populated, the replayed deltas
+    // concatenate onto stale text and the UI renders "How's that?How's that?".
+    // Flush as transcript.done so the client commits its current bubble; post-
+    // resume deltas then start a fresh bubble. Done before the forced-fresh
+    // preamble snapshot so a mid-turn user utterance lands in this.transcript
+    // and gets folded into the rebuilt systemInstruction.
+    this.flushPendingTranscripts()
+
     // Forced-fresh has no resumption handle, so Gemini won't replay any
     // in-call state. Rebuild the systemInstruction preamble from the original
     // conversationHistory PLUS in-call turns accumulated since connect() so
@@ -263,13 +273,6 @@ export class GeminiAdapter implements ProviderAdapter {
     }
     const handlePreview = this.resumptionHandle ? `${this.resumptionHandle.slice(0, 8)}…` : "fresh"
     log(`[gemini] Reconnecting (${reason}, handle=${handlePreview})`)
-    // Finalize any in-flight transcription before rotating. Gemini's resumed
-    // session replays transcription from its last checkpoint, so if we leave
-    // currentAssistantText / currentUserText populated, the replayed deltas
-    // concatenate onto stale text and the UI renders "How's that?How's that?".
-    // Flush as transcript.done so the client commits its current bubble; post-
-    // resume deltas then start a fresh bubble.
-    this.flushPendingTranscripts()
     this.userSpeaking = false
     this.sendToClient?.({ type: "session.rotating" })
     this.pauseWatchdog()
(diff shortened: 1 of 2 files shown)


yagudaev:
## Claude review (parallel to Codex) — different lens

### BLOCKERS
None. The fix is correct, the `HistoryMessage` import resolves (`relay-server/src/history.ts:21` exports it), and Test 3 byte-equality holds because both calls go through the same `rebuildResumePreamble([])` path. Point 3 from the review brief was a false alarm: `reconnect()` already emits `session.rotated` on forced-fresh (`gemini.ts:289`), and `session.ts:119-121` re-pulls `getResumePreamble()`/`getResumeHistory()` into the tracer on that event. The trace will reflect the rebuilt preamble.

### SUGGESTIONS

**[high] Test 1 has a coverage gap on the spec it claims to verify.** The test asserts the resumed `systemInstruction` contains call-start AND in-call terms, but does not assert the *initial* setup (`setupsReceived[0]`) was missing the in-call terms. Add `expect(setupsReceived[0].systemInstruction?.parts?.[0]?.text ?? \"\").not.toContain(\"Four Barrel\")` so a regression where the preamble accidentally accumulates in-call turns at connect time would still fail this test. Without that, the test passes even if the bug fix were a no-op (because rebuilt = original = both contain everything for a tiny history).

**[high] Inlined `mountWithHistory` should be promoted into `mock-gemini.ts`.** The agent's inlining is justified for the immediate ship, but it duplicates ~80 lines of mock plumbing and diverges in subtle ways (no `messagesPerSocket`, no `onMessage` reactor, no `\"never\"` ackSetup). Future PR-260-style tests will keep re-inlining. Refactor `mountMockGemini` to accept an optional `config?: Partial<SessionConfigEvent>` overlay that gets merged onto `TEST_CONFIG` — one-line call site, no duplication. Pre-existing tests need no changes.

**[medium] Behavioral risk worth a live-call check, not a code change.** A new Gemini session with a fresh `systemInstruction` containing turns the model itself \"spoke\" 5-30s ago (now framed as Assistant-said history) is a prompt-engineering edge case. The `formatRecentTurnsPreamble` already says \"Do not re-greet or restate them; pick up naturally\" which mitigates, but please run the manual live test in the test plan before merging — if the model re-greets or restarts, the fix solves the data-loss bug but creates a UX bug.

**[low] Performance is fine — non-issue.** Summarizer only fires above 16 messages (`RECENT_TURNS_VERBATIM * 2`, `history.ts:36-37`). Forced-fresh after 4 initial + 2-4 in-call turns stays well under threshold and skips the LLM call entirely. Even when it does fire, gating it with a fallback to the OLD preamble would silently restore the bug under load — not worth it.

**[low] `getResumeHistory()` not refreshed.** The forced-fresh rebuild updates `resumePreamble` but not whatever `getResumeHistory()` returns. Worth a quick check that `session.ts:121` line `setResumeHistory(resumeHistory)` still sees consistent state on the post-forced-fresh `session.rotated`. Likely fine since `transcript` is the source, but worth a one-line comment.

### Verdict: **ship-with-fixes**

Land the high-priority test assertion strengthening (one-line add) before merge. The mock helper refactor and live behavioral check can land as follow-ups. No blockers.

— Reviewed by Claude (Opus 4.7), parallel to Codex.

PR #266, opened 2026-04-29
fix(brain): persist brain results to local DB independent of model speech

yagudaev:
## Summary

Surface brain agent answers to the desktop client as a first-class `brain.result` relay event so they are persisted into local SQLite the moment the brain returns, independent of whether the realtime model successfully speaks them.

## Failure modes this addresses

1. **Upstream drops mid-injection.** When Gemini closes the WS with 1007 (or any error) during/after `injectContext` but before any audio is generated, the brain's actual answer was lost from local history entirely (we observed this today at 15:46).
2. **Paraphrase fidelity loss on the happy path.** The model's spoken paraphrase is what `transcript.done` captured and persisted, not the brain's full structured response. Cross-session history loaders then fed back a degraded shadow of the structured output to subsequent turns (we saw this today at 14:46 on full-structured brain returns the model summarized poorly).

## Design

Two options were considered:

- **(a) New `brain.result` relay event.** Relay emits the structured answer to the client as soon as `askBrain` resolves; the client persists it directly to local SQLite.
- **(b) Save server-side.** Not viable — the relay is per-user but stateless w.r.t. the desktop's local DB.

This PR ships (a). On success the relay also still calls `injectContext` so the model speaks the answer; the spoken paraphrase still gets persisted via `transcript.done`. By design, two messages may now appear (raw brain + spoken paraphrase): they are different content. `addMessage` handles adjacent same-role rows fine and `getMessages` orders by id, so the cross-call history loader sees them in causal order.

## UX choice (deliberately scope-limited)

The persisted message renders as a normal assistant bubble prefixed with `[Brain] `. No conditional rendering, no settings toggle, no styled component. Visibility / styling polish can land in a follow-up. Choosing visible-by-default keeps the bug fixed even without explicit opt-in.

## Cancellation behavior

When `controller.signal.aborted` is true (user moved on, session replaced, etc.), the existing branches in `handleAskBrain` already discard the `injectContext` and tracer ends with `cancelled`. This PR follows the same convention: **no `brain.result` is emitted on cancellation**, so a tool the user abandoned doesn't clutter history.

- Success path: emit `{ result }` then inject.
- Error path with `!aborted`: emit `{ error }` then inject error inject-context.
- Cancelled (success or error path with `aborted`): skip emit and skip inject.

## Test plan

Relay-side (vitest): `relay-server/test/brain/brain-result-event.test.ts`
- emits `brain.result` on successful response before injecting context
- emits `brain.result` with `error` field on brain failure
- does NOT emit `brain.result` on cancellation

Manual (desktop) — leave to verification, no Electron test harness in repo:
- [ ] Run brain-enabled session, ask a question that triggers `ask_brain`. Confirm `[Brain] ...` bubble appears in chat history before/independent of the model's spoken response.
- [ ] Force a Gemini drop mid-injection (or simulate) and confirm the `[Brain] ...` bubble is still in local SQLite after relaunch.
- [ ] Reload conversation; cross-session history sent to next session includes the structured `[Brain]` content rather than only the model paraphrase.
- [ ] Cancel a long-running brain call (start a new turn while one is in flight) and confirm no `[Brain]` bubble lands.

## Diff

- Server: 4 files (types, session, +new test, brain logic untouched)
- Desktop: 2 files (use-realtime hook surface + ChatPage callback wiring)

### DIFF
--- a/desktop/src/renderer/src/lib/use-realtime.ts
+++ b/desktop/src/renderer/src/lib/use-realtime.ts
@@ -48,6 +48,7 @@ export interface RealtimeCallbacks {
   onTranscriptDone?: (text: string, role: 'user' | 'assistant') => void
   onToolCall?: (callId: string, name: string, args: string) => void
   onToolProgress?: (callId: string, summary: string) => void
+  onBrainResult?: (callId: string, query: string, result?: string, error?: string) => void
   onTurnStarted?: () => void
   onTurnEnded?: () => void
   onSessionReady?: (sessionId: string) => void
@@ -175,6 +176,10 @@ export function useRealtime(callbacks: RealtimeCallbacks): RealtimeControls {
 
         case 'tool.progress':
           cb.onToolProgress?.(data.callId, data.summary)
+          break
+
+        case 'brain.result':
+          cb.onBrainResult?.(data.callId, data.query, data.result, data.error)
           break
 
         case 'turn.started':
--- a/desktop/src/renderer/src/pages/ChatPage.tsx
+++ b/desktop/src/renderer/src/pages/ChatPage.tsx
@@ -173,6 +173,15 @@ export function ChatPage() {
       setStreamingRole('assistant')
       setStreamingText(summary)
     },
+    onBrainResult: async (_callId, query, result, error) => {
+      const body = error
+        ? `[Brain] Failed for "${query}": ${error}`
+        : `[Brain] ${result ?? ''}`
+      if (!body.trim()) return
+      const convId = await ensureConversation()
+      await addMessage(convId, 'assistant', body)
+      await loadMessages()
+    },
     onSessionEnded: () => {
       setIsCallActive(false)
       setIsThinking(false)
--- a/relay-server/src/session.ts
+++ b/relay-server/src/session.ts
@@ -314,7 +314,8 @@ export class RelaySession {
 
       this.tracer.endToolCall(callId, result)
 
-      // Inject the result back into the conversation so Gemini speaks it
+      this.send({ type: "brain.result", callId, query, result })
+
       this.adapter?.injectContext(
         `[Brain agent result for query: "${query}"]\n${result}\n\nPlease share this information with the user naturally.`
       )
@@ -325,6 +326,7 @@ export class RelaySession {
       this.tracer.endToolCall(callId, JSON.stringify({ error: message }), message)
 
       if (!controller.signal.aborted) {
+        this.send({ type: "brain.result", callId, query, error: message })
         this.adapter?.injectContext(
           `[Brain agent failed for query: "${query}": ${message}]\nLet the user know the search didn't work and offer to try again.`
         )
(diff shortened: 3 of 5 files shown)


yagudaev (commented):
Single-reviewer pass (Claude — Codex CLI auth broken this session).

## BLOCKERS
None.

## SUGGESTIONS

**[low] Comment on `BrainResultEvent` (types.ts:233-238) leans bug-history.**
"...isn't degraded to the model's spoken paraphrase across cross-session loads" describes a past failure mode, not an invariant. Per CLAUDE.md the reviewer rule says reject. Borderline though — it doubles as a why-this-event-exists rationale that helps future readers. Tighten to invariant-shaped:

> Raw brain answer surfaced to the client the moment `askBrain` resolves, so the client persists the structured response directly to local history regardless of whether the realtime model later speaks it.

**[low] `result?: string, error?: string` could be a discriminated sub-union.**
`{ result: string, error?: never } | { error: string, result?: never }` would let `if (event.error)` narrow correctly and prevent a future bug emitting both. Not worth blocking — `session.ts` only sets one and the desktop handler treats `error` as the discriminant.

**[low] Test 3 cancellation timing is deterministic, not racy.**
The brain server's handler stashes `respondNow` instead of replying, so the abort fires while the upstream fetch is still pending. The `if (controller.signal.aborted)` check at session.ts:309 fires before line 317's `this.send`. Verified non-flaky.

**[info] `addMessage` ordering is fine.**
`onTranscriptDone` and `onBrainResult` both `await ensureConversation()` then `await addMessage`. ipc-handlers.ts:144 stamps `created_at = Date.now()` and getMessages orders ASC by `created_at`. The two messages are intentionally separate per the PR description; insertion order matches wire-arrival order.

**[info] Exhaustiveness across `switch (data.type)` consumers.**
- `desktop/use-realtime.ts` — handles `brain.result`. Good.
- `mobile/use-realtime.ts:143`, `mobile/chat.ts:229`, `relay-server/test-page.ts:245`, `relay-server/test-e2e.ts:82`, `relay-server/adapters/openai.ts:360` — none have explicit `default` exhaustiveness checks (no `assertNever`); the unknown event is silently ignored, which is the intended forward-compat behavior. Fine.

**[info] Integration with PR #261 graceful shutdown.**
`this.send()` (session.ts:51) gates on `readyState === OPEN` and won't throw if the WS is closing. `brain.result` is emitted synchronously before `injectContext`, so it lands ahead of any close. Safe.

**[info] Cross-session history loader.**
ChatPage `startCall` (line 244-249) loads getMessages → maps to `{role, text}` → sends as `conversationHistory`. Both the `[Brain] ...` row and the spoken paraphrase will be sent on the next session. The Gemini adapter feeds them through `rebuildResumePreamble` (gemini.ts:101). Two adjacent assistant rows with different content is fine — Gemini concatenates them as separate turns; not a double-count, just richer context.

## Verdict: ship

Optional follow-up: tighten the `BrainResultEvent` doc comment to remove the "isn't degraded" history phrasing.

PR #269, opened 2026-04-29
feat(desktop): bundle relay-server as a managed service in DMG (NAN-681 part 1)

yagudaev:
## Summary

Closes the relay half of [NAN-681](https://linear.app/nano3labs/issue/NAN-681) — a fresh-Mac DMG install no longer requires the user to run `yarn dev` or install Node. The desktop app now ships a precompiled relay binary under `Resources/bin/relay-server-darwin-arm64` and spawns it through the existing `service-manager` alongside `openclaw-gateway`.

## Packaging decision: option (b) Bun `--compile`

Three options were on the table:

1. **Node SEA** — official, but tooling is finicky (`postject`, manual `sea-config.json`, separate Node copy) and gives nothing for v1 over Bun.
2. **Bun `--compile`** *(picked)* — one CLI flag, one binary, ~59 MB. Pure-JS deps only (`express`, `ws`, OTel SDKs, `dotenv`) so no native-module surprises. Smoke-tested locally — boots and `/health` returns `{"status":"ok"}`.
3. **Ship Node + `dist/` + `node_modules/`** — ~150 MB, most compatible, but heaviest for zero benefit since Bun handled the deps cleanly.

If Bun ever chokes on a future relay dep, falling back to (3) is a one-script change.

## Dev vs prod

- **Dev (`yarn dev`):** unchanged. The root `concurrently` orchestrator still drives `yarn dev:server` via `tsx watch`. The desktop service-manager spawn is a no-op when the bundled binary is missing (mirrors `openclaw-gateway.ts`).
- **Prod (DMG):** `dist:mac` runs `yarn workspace relay-server build:bin` first, electron-builder copies the result to `Contents/Resources/bin/relay-server-darwin-arm64`, and the main process spawns it on launch.

The renderer's default `ws://localhost:8080/ws` keeps working unchanged because the existing `allocatePort('relay')` in `desktop/src/main/ports.ts` already prefers `8080` with kernel-allocated fallback. No IPC plumbing needed for v1.

## What's deliberately out of scope (next iterations on NAN-681)

- **x64 / universal binary** — v1 ships arm64 only. The `build:bin:x64` script is in place; `lipo`-merging into a universal Mach-O is a follow-up. Documented in the docs page.
- **OpenClaw-gateway binary** — separate scope item; needs openclaw-fork release pipeline work first.
- **First-run wizard for API keys** — v1 forwards `GEMINI_API_KEY`, `OPENAI_API_KEY`, `XAI_API_KEY`, `TAVILY_API_KEY`, `BRAIN_GATEWAY_URL`, `RELAY_API_KEY`, `LANGFUSE_*`, `TRACING_UI_COLLECTOR_URL` from `process.env` when spawning, so dev-set keys still work. Wizard storage comes later.

## Verification

- `yarn workspace relay-server test` — 72 passed / 2 skipped (unchanged from main)
- `yarn workspace relay-server build` (tsc) — clean
- `yarn workspace voiceclaw-desktop typecheck` — clean
- `yarn workspace voiceclaw-desktop test` — 5 new tests for `resolveBundledBinary` (packaged-with-binary, packaged-without, dev-with, dev-without, x64 selection)
- `yarn workspace voiceclaw-desktop build` (electron-vite) — clean
- `electron-builder --mac dir` — packaged successfully; `Resources/bin/relay-server-darwin-arm64` present in the .app bundle
- Bundled binary smoke test: `PORT=18999 ./relay-server-darwin-arm64` → `curl /health` returns `{"status":"ok"}` ✓

Full-DMG / notarize was skipped to keep the iteration tight.

## Test plan

- [ ] Reviewer pulls branch, runs `yarn install && yarn dev` — relay still starts via tsx, desktop app logs `[relay] bundled binary not found; skipping spawn` (expected, no binary in dev resources dir).
- [ ] Reviewer runs `yarn workspace voiceclaw-desktop run dist:mac` (or just `--mac dir`), confirms `dist/mac-arm64/VoiceClaw.app/Contents/Resources/bin/relay-server-darwin-arm64` exists.
- [ ] Reviewer launches the packaged .app, verifies the relay process is spawned (visible in Activity Monitor as a child of VoiceClaw), `~/Library/Logs/VoiceClaw/relay-server.log` exists, and `curl http://127.0.0.1:8080/health` returns ok.


yagudaev (commented):
## Second-reviewer pass (Claude, parallel to Codex) — fresh-Mac install lens

Verdict: **ship-with-fixes.** Architecture is sound; pivot to Node 22 + JS source removed the Bun-bin compatibility risk. But the fresh-Mac UX has a hard blocker around provider-key delivery, and the bundle is carrying a lot of dead weight.

### BLOCKERS

1. **Provider keys never reach a relay process that's already running.** `desktop/src/main/index.ts:167` spawns the relay at `whenReady`. `desktop/src/main/services/relay-server.ts:75-87` reads keys *once* via `getProviderKey()`. `desktop/src/main/ipc-handlers.ts:264-279` saves keys via `setProviderKey` but never restarts the relay. On a brand-new install, the relay boots with no `GEMINI_API_KEY`, the user enters the key in the wizard, hits Start Call, and `relay-server/src/adapters/gemini.ts:116` throws `"GEMINI_API_KEY not set"`. Silent failure (the relay is "running" — health check passes; only the per-call adapter fails). Need either a relay-restart IPC after `provider:validateAndSave`, or change the adapter to read from the keychain on every connect.

2. **`--auth none` + `--allow-unconfigured` on the bundled gateway.** Confirmed bind defaults to `loopback`/127.0.0.1 (`desktop/resources/openclaw/dist/gateway-cli-DN6P_1e8.js:213`), so the network is contained — but anything running on the user's Mac (browser tab, malicious npm package they `node`'d earlier, MCP server) can `fetch http://127.0.0.1:<port>/v1/...` and exfiltrate brain state. Openclaw itself logs `"Gateway auth mode=none explicitly configured; all gateway connections are unauthenticated"` (line 865) — that warning exists for a reason. Pass `--auth token` and either generate a per-install random token in the warmup path or reuse the one openclaw bakes into config after warmup. The relay can read it from `OPENCLAW_GATEWAY_AUTH_TOKEN`.

3. **Doc is stale post-pivot.** `docs/src/content/docs/relay-server.mdx:37-44` still says "ship a precompiled relay binary under `Contents/Resources/bin/relay-server-darwin-arm64`" via "Bun's --compile mode" (~59 MB Mach-O). The PR ships `relay-server-bundle/dist/index.js` invoked through bundled Node 22. Update before merge — these are user-facing docs.

### SUGGESTIONS (high impact)

4. **Disk-size quick wins (~10 MB+ each, no architecture change).**
   - `desktop/resources/openclaw/dist/extensions/browser` = **42 MB** (playwright-core + chromium glue) — only used if user wires up browser extension.
   - `desktop/resources/openclaw/dist/extensions/acpx` = **25 MB** (bundled tsx/esbuild).
   - `desktop/resources/openclaw/dist/extensions/diffs` = **9.5 MB**.
   - Chat platforms users will never use from a desktop Mac app: `slack`, `discord`, `telegram`, `feishu`, `whatsapp`, `qqbot`, `googlechat`, `matrix`, `msteams` — combined ~6 MB.
   - `CHANGELOG.md` = 1.2 MB and `README.md` = 88 KB are already excluded by the `.md` filter in `electron-builder.yml`, but `desktop/scripts/build-openclaw-bundle.mjs` should drop them at staging time so they don't bloat the openclaw subtree before electron-builder runs.

   Add an extension allowlist to `build-openclaw-bundle.mjs` that wipes everything under `dist/extensions/` not in `{ anthropic, anthropic-vertex, google, codex, claude-cli, ... }`. Easy 50–60 MB win.

5. **First-call latency on fresh install ≈ 50–80 s.** Wizard (~1 s) + warmup gateway (up to 30 s; `WARMUP_TIMEOUT_MS` in `openclaw-gateway.ts:9`) + steady-state openclaw spawn + health check (~5–10 s) + relay spawn (~1–2 s) + WS connect + first brain call (10–20 s). This is "I think it's broken" territory. The wizard should:
   - Run the warmup *during* the wizard (not after), so the 30 s overlaps with the user typing keys.
   - Surface a "Warming up local brain (first launch only — ~30 s)" status. Right now the user sees nothing.

6. **Submodule update workflow is undocumented.** `.gitmodules` points at `https://github.com/yagudaev/openclaw.git` (public, so CI clone is fine without a deploy key). But there's no doc / script / Makefile target for "bump openclaw fork to latest" — future me (or another maintainer) will have to remember `git submodule update --remote vendor/openclaw && commit`. Add a one-line script (`yarn workspace desktop run update:openclaw`) or at minimum a section in `relay-server.mdx`.

7. **Warmup port collision.** `runWarmupGateway` picks `40000 + random(1000)` (`openclaw-gateway.ts:112`). If a user double-clicks the dock icon (race vs single-instance lock) or the warmup runs concurrently with a previously-killed leftover, two instances grab the same port. Use `allocatePort` (which already uses kernel auto-allocation) instead of `Math.random()`.

8. **Test thinness.** 7 tests, all `resolveBundled*Script` path-resolution unit tests with `vi.mock('fs')`. Zero integration coverage that actually spawns the bundled `node + openclaw.mjs + relay/dist/index.js` and asserts `/health` returns 200. The agent's manual smoke is the only verification this works end-to-end — and it won't run in CI on every release. Add at least one test that:
   - Runs `node scripts/build-services.mjs` (or asserts the staged dirs exist after a prior build).
   - Spawns the actual bundled relay against the bundled node, hits `/health`, kills it.

9. **`dropBinSymlinks` is overbroad.** `build-openclaw-bundle.mjs:52-74` removes **every** symlink in the tree, including valid ones inside nested `node_modules/.bin/` (verified 7 still surviving after re-run because the recursive subtree's `npm install` re-creates them). If any extension genuinely shells out to a `.bin/` entry at runtime (acpx ships `acpx`, `tsx`, `esbuild`, `skillflag`), this will silently break it. Only the *broken* symlinks (where `existsSync(readlinkSync(p))` is false) need stripping for electron-builder. Verify which extensions actually invoke `.bin/` shims before merging.

### MINOR

10. `desktop/resources/openclaw-config-template.json` hardcodes `gateway.auth.mode = "none"` — this is the seed config for first-launch, so it'd persist across restarts even if the warmup baked a token. Drop the `auth` block from the template; let openclaw bootstrap defaults.
11. `relay-server/scripts/build-bundle.mjs:45` runs `npm install --omit=dev` from a yarn workspace — works, but means two package managers are installed in dev. Document why.
12. CI workflow uses both `corepack enable` (yarn) and `pnpm/action-setup` separately — fine, but `pnpm` is only used for `vendor/openclaw` build. Worth a comment in `release-desktop.yml:64-67`.

### What's good

- Submodule choice (`vendor/openclaw`) is clean — pinned, reproducible, no `npm pack` URL footguns.
- `serviceManager` health-check + crash-detection is solid (200 ms poll, exit-before-healthy detection).
- `node-runtime.ts` packaged-vs-dev resolution is correct.
- Loopback-only bind by default is correct.
- `submodules: recursive` on checkout (`release-desktop.yml:53`) is correct.
- Cert/notarization plumbing untouched and still good.

PR #290, opened 2026-04-30
chore(main): release mobile 1.2.0

yagudaev:
:robot: I have created a release *beep* *boop*
---

### Features

* **mobile:** inline tool-call rows in chat ([#289](https://github.com/yagudaev/voiceclaw/issues/289)) ([c17174b](https://github.com/yagudaev/voiceclaw/commit/c17174b413c48d7bb149afed85f3d47a50ecf522))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #292, opened 2026-04-30
feat(desktop): in-app auto-update via electron-updater (NAN-688)

yagudaev:
## Summary

- Wires `electron-updater` against the existing GitHub Releases publish target — auto-download in background, `autoInstallOnAppQuit: true`, check on launch + every 4 hours
- In-app update banner (Part E) mounts at the top of the app shell when an update is staged; surfaces version, Restart now, Later, and expandable release notes
- Settings card "Updates" (Part D) shows current version, last-checked relative time, staged-update prompt with Restart now button, Check for updates button, and View release notes link
- Tray menu (Part C) always shows greyed current-version row above Quit; adds "Install and Restart: X.Y.Z" item when an update is staged (Littlebird pattern), rebuilt live on staged event
- Three IPC handlers (Part B): `updates:getState`, `updates:checkNow`, `updates:installNow` with install-source tagging (banner | settings | tray)
- All updater state transitions logged to `~/Library/Logs/VoiceClaw/auto-updater.log` (Part F)
- User-facing Starlight docs page at `docs/src/content/docs/desktop/updates.mdx`

## Test plan

- [ ] `yarn workspace voiceclaw-desktop typecheck` — passes clean
- [ ] `yarn workspace voiceclaw-desktop build` — no errors
- [ ] Settings page renders "Updates" card with current version and Check for updates button
- [ ] Tray menu shows `VoiceClaw 0.6.0` greyed row above Quit (no staged item when none staged)
- [ ] On a packaged build receiving an update: banner appears, tray gains Install and Restart item
- [ ] Click Restart now in banner / Settings / tray → all three call `autoUpdater.quitAndInstall`
- [ ] No internet → Check for updates surfaces error message without crash
- [ ] `~/Library/Logs/VoiceClaw/auto-updater.log` captures check + result entries

## Coordination note

This PR only adds a new "Updates" Settings card — it does not touch the Debug section. NAN-692 and NAN-693 add Debug-section items; conflicts are additive and straightforward to resolve if they merge first.


yagudaev:
## Gemini review (local CLI)

Run via `gemini -p` against the PR diff. Codex review couldn't run — local auth token expired (refresh token reused on another machine; needs `codex logout && codex login`).

### 1. Bugs & Logic Issues

- **`desktop/src/main/services/auto-updater.ts:195` — Inconsistent error state on launch.** The initial `checkForUpdates()` call inside `initAutoUpdater` catches errors but only logs them. Unlike the manual `checkForUpdatesNow` (line 101), it does NOT update `state.status` to `'error'` or call `broadcast()`. If the app launches without internet, the UI will indefinitely show "Idle" instead of the actual error state.
- **`desktop/src/main/services/auto-updater.ts:114` — Forced restart kills active sessions.** `updaterInstance.quitAndInstall(false, true)` uses `isForceRunAfter: true`. This triggers an immediate quit. If the user has an active recording or unsaved chat session, this kills the process without triggering `beforeunload` or `will-quit` handlers in a cancellable way. Consider passing `false` for the second arg, or notifying the renderer to save state before the call.
- **`desktop/src/main/services/auto-updater.ts:47` — Broadcast only targets main window.** Uses `getMainWindow()`. If VoiceClaw ever adds a detached Settings/History window, those won't receive update state changes. Safer: `webContents.getAllWebContents().forEach(wc => wc.send(channel, payload))`.

### 2. IPC & Process Boundary

- **`desktop/src/main/ipc-handlers.ts:360` — `updates:installNow` registered via `ipcMain.handle` but returns `void`.** Since `quitAndInstall` terminates the app, the promise never resolves. `ipcMain.on` (fire-and-forget) is more idiomatic for terminal actions.
- **`desktop/src/renderer/src/components/UpdateBanner.tsx:25` — Missing error/state-change listener.** The banner listens for `onStaged` but not `onStateChanged`. If a previously-staged update is revoked or a subsequent check errors, the banner remains visible with stale data.

### 3. Security & Signing

- **Signature verification relies on `electron-builder.yml` config (not in diff).** electron-updater verifies code signature on macOS by default. Confirm `appleId`, `appleIdPassword`, `teamId`, and `publisherName` in the builder config match the cert exactly — otherwise users will see "Update check failed" in production.
- **`desktop/src/main/services/auto-updater.ts:135` — Dynamic `require('electron-updater')`.** Listed in `dependencies` (correct). Verify `electron-vite` treats it as external or properly bundles it; the `try/catch` around `require` may be stripped at build time depending on bundler settings.

### 4. Edge Cases & Error Handling

- **`desktop/src/main/services/auto-updater.ts:60` — DB locking on `persistLastChecked`.** Synchronous `INSERT` while `getDb()` holds a heavy write lock could throw "database is locked". Caught at line 66, but the success won't be recorded.
- **Partial download state not persisted.** `download-progress` percentage is broadcast but not stored in the `state` object. If the user closes/reopens Settings during a download, the progress bar disappears until the next event.
- **`desktop/src/renderer/src/pages/SettingsPage.tsx:774` — `relativeTime` doesn't handle clock skew.** Common after wake-from-sleep. If `ts > Date.now()`, returns "0 days ago" or similar nonsense.

---

None of these are merge-blockers individually, but #1 and #5 are user-visible bugs in the failure paths (no internet on launch, revoked update). Worth a follow-up PR addressing at least those two.

yagudaev:
## Codex review (local CLI)

Run via `codex review --base origin/main` from the PR branch worktree. Three concrete blockers/bugs in the auto-updater that should be fixed before merge.

### P1 — Wrong release feed (BLOCKER, feature won't work on merge)

`desktop/src/main/services/auto-updater.ts:195`

This repo publishes BOTH `desktop-v*` AND `mobile-v*` GitHub releases. The default `electron-updater` GitHub provider resolves the repo-level latest release. **As of right now, `mobile-v1.2.0` is the latest release on this repo** (shipped today via PR #290). After merge, the desktop updater would look for `latest-mac.yml` on that mobile release, hit ERR_UPDATER_LATEST_VERSION_NOT_FOUND, and desktop auto-updates would die until another desktop release becomes the GH-latest.

**Fix:** configure a desktop-specific feed before starting automatic checks. Either:
- Set `setFeedURL({ provider: 'github', owner, repo, releaseType: 'release' })` with a `channel` filter that only matches `desktop-*` tags, OR
- Use the `releaseType: 'release'` option with a custom URL that points at `releases?per_page=100` filtered by `desktop-` prefix, OR
- Switch to a generic feed pointing at a `desktop/latest-mac.yml` published on a custom path.

Verify by manually running the updater in a packaged build with `mobile-v*` as the latest GH release — it must NOT try to fetch from the mobile tag.

### P1 — Restart cancelled by close-to-hide lifecycle

`desktop/src/main/services/auto-updater.ts:109`

`installNow()` calls `updater.quitAndInstall(false, true)` without first setting the app's `quitting` flag. The app's `createMainWindow` already vetoes `close` events unless `markQuitting()` was called (close-to-hide pattern for tray apps). `quitAndInstall()` closes BrowserWindows BEFORE the normal `before-quit` flow, so the main window just hides → restart cancels → user clicks "Restart now" and nothing happens.

**Fix:** call `markQuitting()` (or whatever the existing helper is named) before `quitAndInstall()`. Alternatively, handle `before-quit-for-update` event from electron-updater to set the flag automatically.

### P2 — Status guard breaks staged-install clicks

`desktop/src/main/services/auto-updater.ts:107`

`installNow()` gates execution on `state.status`. But a later periodic check (4h interval) can flip `status` to `'checking'` or `'error'` while `stagedVersion` is still populated and the installer is on disk. The banner / Settings / tray still show "Restart now" buttons, but clicking does nothing.

**Fix:** gate on `state.stagedVersion != null` (or whatever indicates a downloaded update is ready), not on `status`. Or: keep `status` as `'staged'` once an installer is ready, regardless of subsequent check outcomes.

---

These 3 + Gemini's earlier findings (silent launch error, banner missing state-change listener) should all be addressed in a follow-up commit before merging.

yagudaev:
## Review fixes applied

All P1 and P2 findings from the Codex + Gemini reviews have been addressed in a follow-up commit on this branch. The branch has been rebased against main (resolving additive conflicts in `ipc-handlers.ts` and `SettingsPage.tsx`). Typecheck is clean.

### P1 fixes

**Wrong release feed** — switched the updater from the default GitHub provider to a generic provider pointing at `https://raw.githubusercontent.com/yagudaev/voiceclaw/main/desktop/releases/latest-mac.yml`. The release workflow now has an extra step that commits `latest-mac.yml` into `desktop/releases/` on main after each successful desktop release. This ensures the updater always resolves the latest desktop version, regardless of what GitHub considers the repo-level latest release.

**Restart cancelled by close-to-hide** — `markQuitting()` is now called before `quitAndInstall()` in `installNow()`. The existing close-veto handler in `window-lifecycle.ts` already checks that flag.

**Status guard breaks staged-install clicks** — `installNow()` now gates on `state.stagedVersion != null` instead of `state.status === 'staged'`. A subsequent background check (checking / error) can no longer silently block a pending install.

### P2 fixes

**Silent launch-time errors** — the initial `checkForUpdates()` call in `initAutoUpdater` now mirrors the full error handling from `checkForUpdatesNow()`: sets `state.status = 'error'`, timestamps, persists, and broadcasts `updates:stateChanged`.

**Banner missing state-change listener** — `UpdateBanner` now subscribes to `onStateChanged` in addition to `onStaged`, clearing or updating the banner when a new check revokes or replaces a staged update.

### Deferred (not blockers — noted for follow-up)

- `broadcast()` targets only the main window; detached windows won't receive update events. Deferred since VoiceClaw currently has a single renderer window.
- `ipcMain.on` vs `ipcMain.handle` for `updates:installNow` — idiomatic difference only, not functional.
- DB locking on `persistLastChecked` — already caught and logged; acceptable for a metrics-only write.
- Partial download progress not persisted to state between restarts — low impact since `autoInstallOnAppQuit` handles the common path.
- Relative time clock skew — cosmetic only.

### Hold on merge

PR is not being merged yet — waiting for a hotfix on the `archiver` packaging regression (v0.7.0+) to land first so release-please doesn't cut another broken DMG.

PR #294, opened 2026-04-30
feat(desktop): show/hide toggle on onboarding API key field (NAN-683)

yagudaev:
## Why
- Onboarding masks the API key, but users have no way to verify what they pasted — they either retype blindly or delete and start over.
- A standard eye-toggle is the lowest-friction fix and keeps masked-by-default for shoulder-surfing safety.

## What
- `StepProvider.tsx`: added `showKey` state (default `false`) and an `Eye`/`EyeOff` button (`lucide-react`) absolutely positioned at the right edge of the API key input.
- Switched input `type` to `showKey ? 'text' : 'password'`.
- Reset `showKey` when the user changes provider (the same place we reset `apiKey`).
- Tweaked input padding (`pr-14`) so the key text never collides with the toggle.
- `aria-label` flips between `Show API key` / `Hide API key`; `aria-pressed` reflects state; `type="button"` so it doesn't submit.
- No changes to validation, the `Continue` button gating, or any other onboarding step.

## Screenshots
Reference from the ticket:

ticket reference

Before/after captures of the running UI weren't taken — the component lives in the Electron renderer with a preload-bridged onboarding API, so it's not trivially bootable in headless Playwright. Verified via `yarn workspace voiceclaw-desktop typecheck` + `build`. Worth a manual eyeball in `yarn workspace voiceclaw-desktop dev` before merging.

## Test plan
- [ ] `yarn workspace voiceclaw-desktop dev`, walk to the provider step
- [ ] Field starts masked; eye icon visible at right edge
- [ ] Click eye → key reveals as plaintext, icon flips to `EyeOff`
- [ ] Click again → re-masks, icon flips back to `Eye`
- [ ] Switch provider (Gemini ↔ Grok) → input clears AND re-masks
- [ ] Paste a key, tab out → still validates and shows the existing badge
- [ ] `Continue` flow unchanged (validate → green check → continue)
- [ ] Keyboard: Tab focuses the toggle, Space/Enter activates it
- [ ] Screen reader announces `Show API key` / `Hide API key`

### DIFF
--- a/desktop/src/renderer/src/pages/onboarding/StepProvider.tsx
+++ b/desktop/src/renderer/src/pages/onboarding/StepProvider.tsx
@@ -217,7 +217,7 @@ export function StepProvider({
               onClick={() => setShowKey((prev) => !prev)}
               aria-label={showKey ? 'Hide API key' : 'Show API key'}
               aria-pressed={showKey}
-              className="absolute right-3 top-1/2 flex h-8 w-8 -translate-y-1/2 items-center justify-center rounded-[10px] transition-colors hover:bg-[var(--panel)] focus:bg-[var(--panel)] focus:outline-none"
+              className="absolute right-3 top-1/2 flex h-8 w-8 -translate-y-1/2 items-center justify-center rounded-[10px] transition-colors hover:bg-[var(--panel)] focus:bg-[var(--panel)] focus:outline-none focus-visible:ring-2 focus-visible:ring-offset-1 focus-visible:ring-[var(--ink)]"
               style={{ color: 'var(--muted)' }}
             >
               {showKey ? (
--- a/desktop/src/renderer/src/pages/onboarding/StepProvider.tsx
+++ b/desktop/src/renderer/src/pages/onboarding/StepProvider.tsx
@@ -1,4 +1,5 @@
 import { useState } from 'react'
+import { Eye, EyeOff } from 'lucide-react'
 import { StepFrame } from './StepFrame'
 import {
   providerApi,
@@ -69,6 +70,7 @@ export function StepProvider({
     PROVIDERS.some((p) => p.id === initialProvider) ? initialProvider : 'gemini',
   )
   const [apiKey, setApiKey] = useState('')
+  const [showKey, setShowKey] = useState(false)
   const [validation, setValidation] = useState<ValidationState>({ kind: 'empty' })
   const active = PROVIDERS.find((p) => p.id === selected) ?? PROVIDERS[0]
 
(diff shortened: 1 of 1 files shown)


desktop/src/renderer/src/pages/onboarding/StepProvider.tsx
```diff
@@ -209,6 +212,20 @@ export function StepProvider({
                 letterSpacing: '0.04em',
               }}
             />
+            <button
+              type="button"
+              onClick={() => setShowKey((prev) => !prev)}
+              aria-label={showKey ? 'Hide API key' : 'Show API key'}
+              aria-pressed={showKey}
+              className="absolute right-3 top-1/2 flex h-8 w-8 -translate-y-1/2 items-center justify-center rounded-[10px] transition-colors hover:bg-[var(--panel)] focus:bg-[var(--panel)] focus:outline-none"
```
yagudaev: The button uses `focus:outline-none` without a `focus-visible:ring-*` replacement, leaving keyboard users with no visible focus indicator on this control. WCAG 2.4.7 (Focus Visible) requires a visible focus state. A minimal fix:

```tsx
className="absolute right-3 top-1/2 flex h-8 w-8 -translate-y-1/2 items-center justify-center rounded-[10px] transition-colors hover:bg-[var(--panel)] focus:outline-none focus-visible:ring-2 focus-visible:ring-[var(--accent)] focus-visible:ring-offset-1"
```

The background color change on `focus:` alone is insufficient — colour-only indicators fail WCAG 1.4.1 and are invisible to users who can't distinguish subtle hue differences.

yagudaev (commented):
One blocking a11y issue to fix before merging.

**Focus ring missing on eye toggle (line 220)**
The button has `focus:outline-none` with no `focus-visible:ring-*` fallback. Keyboard users tabbing to the toggle get zero visible focus indicator. This violates WCAG 2.4.7 (Focus Visible) and 1.4.1 (Use of Color) — a background color shift alone isn't sufficient. Fix: swap to `focus:outline-none focus-visible:ring-2 focus-visible:ring-[var(--accent)] focus-visible:ring-offset-1`.

Everything else is clean:
- No semicolons, helpers at bottom — style rules pass.
- `showKey` resets correctly on provider switch (line 152).
- `type="button"`, `aria-label` flip, and `aria-pressed` are all correct.
- Validation flow, Continue button gating, and `onBlur` trigger are untouched.
- `pr-14` padding prevents text/icon collision.
- One inline comment posted at line 220 with the exact fix.

yagudaev:
Addressed in efe526c: added `focus-visible:ring-2 focus-visible:ring-offset-1 focus-visible:ring-[var(--ink)]` on the eye toggle. Typecheck + build clean.

PR #296, opened 2026-04-30
feat(desktop): right-click delete message in chat (NAN-679)

yagudaev:
## Why

Users have no way to remove a message from the desktop chat — bad turns (mis-transcribed user audio, off-topic assistant replies, broken tool-call narration) accumulate in the persisted history and get re-sent as conversation context on the next call.

NAN-679 ships the smallest fix: right-click a bubble, pick **Delete message**, the row disappears from the UI and from `voiceclaw.db`.

## What

- New renderer component `MessageContextMenu.tsx` — small, dependency-free, anchored at the cursor with overflow clamping; closes on outside click, Escape, scroll, resize.
- `MessageBubble` accepts an optional `onContextMenu(event, message)` prop and calls `e.preventDefault()` so the OS menu never appears.
- `ChatPage` owns the open menu state and the delete/copy handlers. Delete prompts via `window.confirm` (since the row is gone from sqlite), optimistically removes from local state, and re-loads on IPC failure.
- New IPC channel **`db:deleteMessage`** in `ipc-handlers.ts` — scoped to a single id (`DELETE FROM messages WHERE id = ?`), validates the id is a positive number, returns `{ ok: false, error }` for missing rows. Bumps `conversations.updated_at`. No bulk delete.
- Menu items: **Copy** (writes content to clipboard) and **Delete message** (destructive style).

Honors the project's no-data-wipe rule: deletion is one row, by id, on explicit user action with a confirmation prompt.

## Test plan

- [ ] `yarn workspace voiceclaw-desktop typecheck` — passes
- [ ] `yarn workspace voiceclaw-desktop build` — passes
- [ ] `yarn workspace voiceclaw-desktop test` — 39/39 pass
- [ ] Manual: start `yarn workspace voiceclaw-desktop dev`, run a short call so a few user + assistant + tool-call rows exist
- [ ] Right-click a user bubble — menu appears at cursor with Copy + Delete; OS menu does not show
- [ ] Click **Delete message** — confirm dialog appears, accept, bubble disappears immediately
- [ ] Cmd+R (reload renderer) — deleted message stays gone (sqlite was updated)
- [ ] Right-click an assistant bubble — same flow works
- [ ] Right-click on a streaming/in-progress assistant bubble — only persisted (committed) bubbles get the handler; the live streaming preview has no menu (it isn't a `Message`)
- [ ] Tool-call rows have no context menu (intentional — they're not in `messages` table)
- [ ] Delete the only remaining message — UI shows empty state cleanly
- [ ] Esc / outside-click / scroll closes the menu without selecting

## Screenshots

Manual visual verification only — no Playwright snapshot in this PR. Will attach a screen recording on request.

## Caveats

- IPC name uses the existing `db:*` namespace to match `db:deleteConversation`.
- Confirmation uses `window.confirm`; happy to swap for a styled `AlertDialog` once one lands in `components/ui/`.

### DIFF
--- a/desktop/src/renderer/src/components/MessageContextMenu.tsx
+++ b/desktop/src/renderer/src/components/MessageContextMenu.tsx
@@ -1,0 +1,86 @@
+import { useEffect, useLayoutEffect, useRef, useState, type ReactNode } from 'react'
+
+export type MessageContextMenuItem = {
+  label: string
+  onSelect: () => void
+  destructive?: boolean
+  icon?: ReactNode
+}
+
+interface MessageContextMenuProps {
+  x: number
+  y: number
+  items: MessageContextMenuItem[]
+  onClose: () => void
+}
+
+const MENU_MARGIN = 8
+
+export function MessageContextMenu({ x, y, items, onClose }: MessageContextMenuProps) {
+  const ref = useRef<HTMLDivElement>(null)
+  const [position, setPosition] = useState({ x, y })
+
+  useLayoutEffect(() => {
+    const el = ref.current
+    if (!el) return
+    const rect = el.getBoundingClientRect()
+    let nextX = x
+    let nextY = y
+    if (nextX + rect.width + MENU_MARGIN > window.innerWidth) {
+      nextX = Math.max(MENU_MARGIN, window.innerWidth - rect.width - MENU_MARGIN)
+    }
+    if (nextY + rect.height + MENU_MARGIN > window.innerHeight) {
+      nextY = Math.max(MENU_MARGIN, window.innerHeight - rect.height - MENU_MARGIN)
+    }
+    setPosition({ x: nextX, y: nextY })
+  }, [x, y])
+
+  useEffect(() => {
+    const onPointerDown = (e: PointerEvent) => {
+      if (ref.current && !ref.current.contains(e.target as Node)) onClose()
+    }
+    const onKey = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') onClose()
+    }
+    const onScroll = () => onClose()
+    window.addEventListener('pointerdown', onPointerDown, true)
+    window.addEventListener('keydown', onKey)
+    window.addEventListener('resize', onScroll)
+    window.addEventListener('scroll', onScroll, true)
+    return () => {
+      window.removeEventListener('pointerdown', onPointerDown, true)
+      window.removeEventListener('keydown', onKey)
+      window.removeEventListener('resize', onScroll)
+      window.removeEventListener('scroll', onScroll, true)
+    }
+  }, [onClose])
+
+  return (
+    <div
+      ref={ref}
+      role="menu"
+      style={{ top: position.y, left: position.x }}
+      className="fixed z-50 min-w-[180px] rounded-md border border-border bg-card text-foreground vc-panel-shadow py-1"
+    >
+      {items.map((item, i) => (
+        <button
+          key={i}
+          role="menuitem"
+          onClick={() => {
+            item.onSelect()
+            onClose()
+          }}
+          className={
+            'w-full flex items-center gap-2 text-left px-3 py-1.5 text-sm transition-colors ' +
+            (item.destructive
+              ? 'text-destructive hover:bg-destructive/10'
+              : 'text-foreground hover:bg-accent hover:text-accent-foreground')
+          }
+        >
+          {item.icon && <span className="shrink-0">{item.icon}</span>}
+          <span>{item.label}</span>
+        </button>
+      ))}
+    </div>
+  )
+}
--- a/desktop/src/renderer/src/components/MessageContextMenu.tsx
+++ b/desktop/src/renderer/src/components/MessageContextMenu.tsx
@@ -18,6 +18,7 @@ const MENU_MARGIN = 8
 
 export function MessageContextMenu({ x, y, items, onClose }: MessageContextMenuProps) {
   const ref = useRef<HTMLDivElement>(null)
+  const previousFocusRef = useRef<HTMLElement | null>(null)
   const [position, setPosition] = useState({ x, y })
 
   useLayoutEffect(() => {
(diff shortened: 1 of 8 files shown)


desktop/src/renderer/src/components/MessageContextMenu.tsx
```diff
@@ -0,0 +1,86 @@
...
+    let nextX = x
+    let nextY = y
+    if (nextX + rect.width + MENU_MARGIN > window.innerWidth) {
+      nextX = Math.max(MENU_MARGIN, window.innerWidth - rect.width - MENU_MARGIN)
+    }
+    if (nextY + rect.height + MENU_MARGIN > window.innerHeight) {
+      nextY = Math.max(MENU_MARGIN, window.innerHeight - rect.height - MENU_MARGIN)
+    }
+    setPosition({ x: nextX, y: nextY })
+  }, [x, y])
```
yagudaev, line 37: **Accessibility: missing focus management and keyboard navigation**

The `<div role="menu">` is rendered into the DOM but no element receives focus when the menu opens, and individual `<button role="menuitem">` elements don't support arrow-key navigation.

The ARIA menu pattern (W3C APG) requires:
1. Focus moves to the first (or last) `menuitem` when the menu opens.
2. Arrow Up/Down moves focus between items.
3. Focus returns to the triggering element on dismiss.

Without (1), keyboard users (and screen readers) have no way to operate the menu after it appears — even though it's correctly marked up with `role="menu"`. The Escape-key listener fires, but focus never entered the menu in the first place.

Minimal fix: call `ref.current?.querySelector<HTMLElement>('[role=menuitem]')?.focus()` at the end of the `useLayoutEffect`, and store + restore the activeElement on close.

desktop/src/renderer/src/pages/ChatPage.tsx
```diff
@@ -74,6 +77,41 @@ export function ChatPage() {
...
+    },
+    [],
+  )
+
+  const closeMessageMenu = useCallback(() => setMessageMenu(null), [])
+
+  const handleDeleteMessage = useCallback(async (message: Message) => {
+    const ok = window.confirm(
+      'Delete this message? This removes it from the conversation history and cannot be undone.',
+    )
```
yagudaev, line 93: **window.confirm blocks the renderer, but a more serious issue: the delete handler calls confirm() while the menu is still mounted**

`window.confirm` is synchronous and works in Electron (no suppression detected in the window config), so it won't silently no-op. However, calling it from inside `handleDeleteMessage` while `messageMenu` is still non-null means the menu div remains visible underneath the native dialog. In practice this is a visual artifact, not a correctness bug, but closing the menu before the confirm prompt would be cleaner:

```ts
closeMessageMenu()
const ok = window.confirm('…')
```

Also: the PR body itself says "happy to swap for a styled AlertDialog once one lands in components/ui/" — that's the right destination; just noting the visual glitch in the meantime.

desktop/src/renderer/src/components/MessageContextMenu.tsx
```diff
@@ -0,0 +1,86 @@
...
+              : 'text-foreground hover:bg-accent hover:text-accent-foreground')
+          }
+        >
+          {item.icon && <span className="shrink-0">{item.icon}</span>}
+          <span>{item.label}</span>
+        </button>
+      ))}
+    </div>
+  )
+}
```
yagudaev, line 114: **Missing docs page for a user-visible feature**

Per the project rule (feedback_ship_docs_with_features): every UI-facing feature ships with a Starlight docs page in `docs/src/content/docs/`. Right-click delete is user-facing and not covered here. Please add a short page (or extend an existing "chat" page) before merge.

yagudaev (commented):
## Review — REQUEST CHANGES (posted as comment; GitHub blocks self-review as request-changes)

### Verdict: needs fixes before merge — no blockers on data-safety or SQL, but two required items and one nit.

---

### SQL / data safety — CLEAN

- DELETE is parameterized (`WHERE id = ?`), single-row, validated to a positive finite number. Satisfies `feedback_no_data_wipe`.
- No dependent tables reference `messages` — the schema has no `tool_calls` or `attachments` table, so there are no orphan risks.
- `foreign_keys = ON` pragma is set at DB open; the `conversations` bump is a nice touch.
- Optimistic rollback via `loadMessages()` is correct and restores order.

---

### Required fixes

**1. Accessibility — focus not moved into menu on open (inline comment, line 36 of MessageContextMenu.tsx)**
The `role="menu"` + `role="menuitem"` markup is correct but no element receives focus when the menu mounts. Keyboard users cannot Tab into it, and Shift-F10 opens the menu via the browser default but focus stays on the triggering bubble. Add `ref.current?.querySelector<HTMLElement>('[role=menuitem]')?.focus()` at the end of the `useLayoutEffect`, and store/restore `document.activeElement` on open/close.

**2. Missing Starlight docs page (inline comment, line 86 of MessageContextMenu.tsx)**
Project rule (`feedback_ship_docs_with_features`): every user-visible feature ships a docs page in `docs/src/content/docs/`. Right-click delete is user-facing and needs one. Even a short addition to an existing "Using the chat" page is enough.

---

### Nit (non-blocking)

**3. menu stays visible under window.confirm (inline comment, ChatPage.tsx line 92)**
Call `closeMessageMenu()` before `window.confirm()` to avoid the menu painting behind the native dialog. No correctness risk — just a visual artifact.

---

### Things reviewed and found clean

- No semicolons, no banned ticket-ID or bug-history source comments.
- `buildMessageMenuItems` helper is at the bottom of ChatPage — correct file organization.
- Symlinks (`.yarn`, `desktop/node_modules`) are untracked; no worktree pollution committed.
- Viewport clamping logic is straightforward, not over-engineered.
- Outside-click, Escape, scroll, and resize all close the menu.
- Streaming preview is rendered separately (not via `MessageBubble`) so it cannot receive a context menu.
- IPC name follows existing `db:*` namespace convention.

yagudaev (commented):
## Review — needs fixes before merge

### Verdict: two required items, one nit. No blockers on SQL/data-safety.

---

### SQL / data safety — CLEAN

- DELETE is parameterized, single-row, positive-int validated. Satisfies `feedback_no_data_wipe`.
- No dependent tables (`tool_calls`, `attachments`) reference `messages` in the schema — no orphan risk.
- Optimistic rollback via `loadMessages()` on IPC failure is correct and order-preserving.

---

### Required fixes

**1. Accessibility — focus not moved into menu on open** (inline comment on MessageContextMenu.tsx:36)
`role="menu"` + `role="menuitem"` markup is correct, but no element receives focus when the menu mounts. Keyboard users cannot reach the items, and Shift-F10 leaves focus on the triggering bubble. Fix: `ref.current?.querySelector<HTMLElement>('[role=menuitem]')?.focus()` at the end of `useLayoutEffect`, plus store/restore `document.activeElement` on open/close.

**2. Missing Starlight docs page** (inline comment on MessageContextMenu.tsx:86)
Project rule (`feedback_ship_docs_with_features`): every user-facing feature ships a docs page in `docs/src/content/docs/`. Right-click delete needs one — even a short addition to an existing chat page.

---

### Nit

**3. Menu stays visible under window.confirm** (inline comment on ChatPage.tsx:92)
Call `closeMessageMenu()` before `window.confirm()` to prevent the menu painting behind the native dialog. No correctness risk.

---

### Found clean

- No semicolons; no NAN-679 or bug-history source comments; `buildMessageMenuItems` helper is at the bottom of ChatPage.
- Symlinks (`.yarn`, `desktop/node_modules`) are untracked — no worktree pollution.
- Viewport clamping is simple and correct. Outside-click, Escape, scroll, and resize all close the menu.
- Streaming preview is not a `MessageBubble` and cannot receive a context menu — the guarding is correct.

yagudaev:
Addressed all three review items in 3c24a77:

1. **Focus management** — `MessageContextMenu` now stores the previous `activeElement` on mount, focuses the first `[role=menuitem]` on open, and restores focus on unmount. Added Arrow Up/Down/Home/End navigation between items.
2. **Docs page** — added `docs/src/content/docs/desktop/chat-actions.mdx` (user-facing copy, no implementer specs) and wired into the Guides sidebar.
3. **Close-before-confirm** — `setMessageMenu(null)` runs before `window.confirm` in `ChatPage.handleDeleteMessage` so the menu no longer paints behind the dialog.

Typecheck, build, and 39/39 vitest cases all pass.

Note: I did not add Shift-F10 keyboard activation in this PR — that requires making message bubbles tab-focusable, which is broader scope. Filed mentally as a follow-up; happy to spin up a separate ticket if you want it sooner.

PR #300, opened 2026-04-30
fix(desktop): per-provider voice persistence + Grok default Ara (NAN-676)

yagudaev:
Closes NAN-676.

## Why
Voice was stored as one global string, so the moment a user switched from Gemini to Grok in Settings, their previous voice pick on the other provider was gone. Coming back to Gemini didn't restore Zephyr (or whatever they'd picked), it just reset to a default. The ticket also calls out that Grok should default to Ara instead of the previous fallback (`eve`), since Ara is the preferred sound.

## What
- New persisted setting `realtime_voices_by_provider` — a JSON map `{ gemini, xai }` keyed by provider id.
- Settings UI loads the per-provider value on mount. Picking a model swaps the voice grid to that provider's catalog and restores the last voice for that provider (or the provider default).
- Picking a voice writes to the per-provider map (and mirrors to the legacy `realtime_voice` key so the existing relay code path stays unchanged).
- ChatPage `startCall` now reads voice from `getVoiceForProvider(providerForModel(model))` instead of the single global key.
- IPC `identity:save` (which used to write a single voice string) updates the per-provider entry that matches the saved voice's catalog (or falls back to the active provider).
- Defaults are now centralized: `gemini → Zephyr`, `xai → ara` (lowercase, matches the relay test page + the in-app voice catalog id).
- Shared helpers live in `desktop/src/renderer/src/lib/voice-prefs.ts` (renderer) and `desktop/src/main/voice-prefs.ts` (main). Removes the duplicated `GEMINI_VOICES` / `XAI_VOICES` / `normalizeRealtimeVoice` constants that were copy-pasted across `SettingsPage.tsx` and `ChatPage.tsx`.

## Migration
- First read of `realtime_voices_by_provider` seeds the map from the legacy `realtime_voice` value, slotting it into whichever provider's catalog matches. After that, the legacy key is just a write-through mirror — no destructive change to existing rows.
- Existing users who only ever used Gemini keep their Zephyr/whatever; existing Grok users (likely on `eve`) keep `eve` until they change it; new/unset Grok users now land on `ara`.

## Screenshots
Reference from the ticket:
voice settings

## Test plan
- [x] `yarn workspace voiceclaw-desktop typecheck` — clean.
- [x] `yarn workspace voiceclaw-desktop build` — clean.
- [x] `yarn workspace voiceclaw-desktop test` — 13 new tests for `voice-prefs` (provider mapping, defaults, round-trip per provider, malformed JSON, legacy migration idempotency). 48/48 desktop tests now pass that did before; the one pre-existing `diagnostic-bundle.test.ts` failure on `origin/main` is unrelated.
- [ ] Manual: launch desktop, pick Gemini → Aoede, switch to Grok → expect Ara (default), pick Rex, switch back to Gemini → expect Aoede restored.
- [ ] Manual: existing user with only `realtime_voice = Zephyr` upgrades; open Settings — voice still says Zephyr; switch to Grok — defaults to Ara without losing Zephyr.

### DIFF
--- a/desktop/src/main/voice-prefs.ts
+++ b/desktop/src/main/voice-prefs.ts
@@ -1,0 +1,73 @@
+// Main-process mirror of the renderer's per-provider voice preferences.
+// Used by IPC handlers that need to keep the persisted voice map in sync
+// with state owned in main (e.g. agent identity).
+
+import { getDb } from './db'
+
+export const GEMINI_VOICES = [
+  'Puck',
+  'Charon',
+  'Kore',
+  'Fenrir',
+  'Aoede',
+  'Leda',
+  'Orus',
+  'Zephyr',
+] as const
+
+export const XAI_VOICES = ['eve', 'ara', 'rex', 'sal', 'leo'] as const
+
+export type ProviderId = 'gemini' | 'xai'
+
+export const VOICES_BY_PROVIDER_KEY = 'realtime_voices_by_provider'
+export const LEGACY_VOICE_KEY = 'realtime_voice'
+export const ACTIVE_MODEL_KEY = 'realtime_model'
+
+export function providerForModel(model: string | null | undefined): ProviderId {
+  if (model && model.startsWith('grok-voice-')) return 'xai'
+  return 'gemini'
+}
+
+export function providerForVoice(voice: string): ProviderId | null {
+  if ((GEMINI_VOICES as readonly string[]).includes(voice)) return 'gemini'
+  if ((XAI_VOICES as readonly string[]).includes(voice)) return 'xai'
+  return null
+}
+
+export function setVoiceForProviderSync(provider: ProviderId, voice: string): void {
+  const db = getDb()
+  const upsert = db.prepare(
+    'INSERT INTO settings (key, value) VALUES (?, ?) ON CONFLICT(key) DO UPDATE SET value = ?',
+  )
+  const row = db
+    .prepare('SELECT value FROM settings WHERE key = ?')
+    .get(VOICES_BY_PROVIDER_KEY) as { value: string } | undefined
+  const map = parseVoicesByProvider(row?.value)
+  map[provider] = voice
+  const next = JSON.stringify(map)
+  upsert.run(VOICES_BY_PROVIDER_KEY, next, next)
+  upsert.run(LEGACY_VOICE_KEY, voice, voice)
+}
+
+export function getActiveProvider(): ProviderId {
+  const db = getDb()
+  const row = db
+    .prepare('SELECT value FROM settings WHERE key = ?')
+    .get(ACTIVE_MODEL_KEY) as { value: string } | undefined
+  return providerForModel(row?.value ?? null)
+}
+
+function parseVoicesByProvider(raw: string | null | undefined): Record<string, string> {
+  if (!raw) return {}
+  try {
+    const parsed = JSON.parse(raw) as unknown
+    if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) return {}
+    const out: Record<string, string> = {}
+    for (const [k, v] of Object.entries(parsed)) {
+      if (typeof v === 'string') out[k] = v
+    }
+    return out
+  } catch {
+    return {}
+  }
+}
--- a/desktop/src/main/voice-prefs.ts
+++ b/desktop/src/main/voice-prefs.ts
@@ -1,7 +1,3 @@
-// Main-process mirror of the renderer's per-provider voice preferences.
-// Used by IPC handlers that need to keep the persisted voice map in sync
-// with state owned in main (e.g. agent identity).
-
 import { getDb } from './db'
 
 export const GEMINI_VOICES = [
--- a/desktop/src/renderer/src/lib/voice-prefs.ts
+++ b/desktop/src/renderer/src/lib/voice-prefs.ts
@@ -1,0 +1,97 @@
+// Per-provider voice preferences. Voice selection is persisted as a
+// JSON map keyed by provider so switching models in Settings restores
+// the user's prior voice for that provider instead of dropping it.
+
+import { getSetting, setSetting } from './db'
+
+export const GEMINI_VOICES = [
+  'Puck',
+  'Charon',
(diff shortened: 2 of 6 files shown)


desktop/src/renderer/src/lib/voice-prefs.ts
```diff
@@ -0,0 +1,97 @@
+// Per-provider voice preferences. Voice selection is persisted as a
+// JSON map keyed by provider so switching models in Settings restores
+// the user's prior voice for that provider instead of dropping it.
```
yagudaev: **BLOCKER (CLAUDE.md — comment rule):** The file-level block comment restates what the code does rather than explaining a non-obvious WHY. Per CLAUDE.md: "Don't restate what the code does. Well-named identifiers do that." The phrase "Voice selection is persisted as a JSON map keyed by provider so switching models in Settings restores the user's prior voice" reads like a PR description paragraph, not a hidden constraint.

Remove lines 1–3 entirely. The file name `voice-prefs.ts` and the exported symbol names make the purpose clear without prose.

desktop/src/main/voice-prefs.ts
```diff
@@ -0,0 +1,73 @@
+// Main-process mirror of the renderer's per-provider voice preferences.
+// Used by IPC handlers that need to keep the persisted voice map in sync
+// with state owned in main (e.g. agent identity).
```
yagudaev: **BLOCKER (CLAUDE.md — comment rule):** Same issue as the renderer mirror. This file-level comment restates what the code does and why the file exists in conversational prose. The duplication between renderer and main-process `voice-prefs.ts` is architecturally self-documenting (different runtimes, sync vs async DB APIs, IPC boundary). Drop lines 1–3.

desktop/src/renderer/src/pages/SettingsPage.tsx
```diff
@@ -198,24 +202,21 @@ export function SettingsPage() {
...
-      setVoice('eve')
-      save('realtime_voice', 'eve')
-    }
+    const nextProvider = providerForModel(v)
+    if (isVoiceForProvider(nextProvider, voice)) return
+    void (async () => {
+      const restored = await getVoiceForProvider(nextProvider)
+      setVoice(restored)
+      if (loadedRef.current) await setVoiceForProvider(nextProvider, restored)
+    })()
```
yagudaev, line 211: **Race condition (non-blocking, bounded):** Fast double-click on model toggle can produce a stale UI state: the first async `getVoiceForProvider` resolves after the model state has already flipped back, leaving the voice display showing a voice from the wrong provider's catalog. The active call path is unaffected (`ChatPage` always re-reads `getVoiceForProvider(providerForModel(model))` at call start), and the `isVoiceForProvider` guard on read prevents the stale value from persisting to the call. No data corruption; auto-heals on next Settings load.

Low priority to fix but mentioning it. One option: store a local ref for the in-flight provider and ignore the result if it no longer matches.

yagudaev (commented):
## Verdict: Request Changes (posted as comment — PR is by same author so --request-changes is blocked)

Two BLOCKERs (comment rule from CLAUDE.md) + one non-blocking race condition note. Logic and migration are correct.

---

### BLOCKERS

**File-level block comments in both `voice-prefs.ts` files** restate what the code does rather than explaining a non-obvious WHY constraint. Per CLAUDE.md: "Don't restate what the code does. Well-named identifiers do that." and "Reviewers: reject comments that name a ticket ID, describe a bug or its history, or read like a commit-message paragraph. Treat as BLOCKER." The prose in both files reads like a PR description pasted into source. Drop all three lines from each file. (Inline comments posted.)

---

### Non-blocking findings

**Race in `updateModel`:** Fast double-click can leave the Settings voice display showing the wrong provider's voice after the stale async resolver lands. Bounded — `ChatPage` always re-reads voice from the correct provider at call start, and `isVoiceForProvider` discards cross-catalog values on read. Self-heals on next Settings load. Noted inline; not a blocker for merge.

---

### Clean

- Migration correctness: `seedFromLegacy` correctly classifies by catalog (Zephyr → gemini, rex → xai), not by current model. A user who had `realtime_voice=Zephyr` on Grok gets `gemini: Zephyr`, not `grok: Zephyr`. Correct.
- Idempotency: once `realtime_voices_by_provider` is written, `loadVoicesByProvider` returns it directly without re-seeding. Confirmed by test case "only seeds once".
- Malformed JSON: `parseVoicesByProvider` catches and returns `null`; falls through to legacy migration, writes a clean value. Self-healing.
- Dual-write: both `setVoiceForProvider` (renderer) and `setVoiceForProviderSync` (main) write to legacy `realtime_voice` after updating the map. Consistent.
- Casing: `ara` (lowercase) confirmed in relay-server test page catalog (`relay-server/src/test-page.ts` line 79). Matches.
- Helper placement: `parseVoicesByProvider` and `seedFromLegacy` correctly sit at the bottom of the renderer file. Main-process file does the same.
- No semicolons anywhere in the new files.
- Mobile isolation: mobile reads only `realtime_voice` (the legacy key). Desktop dual-writes to that key so mobile continues to work. No shared schema breakage.
- Duplication (renderer + main): necessary. Renderer uses async `getSetting`/`setSetting` over IPC; main uses synchronous better-sqlite3. The only alternative would be an additional IPC round-trip from main back into renderer, which is worse. Duplication is acceptable and the files stay small.
- Tests: 13 cases cover defaults, round-trips, provider isolation, catalog mismatch fallback, all three migration paths (Gemini legacy, Grok legacy, idempotency), and malformed JSON. Meaningful coverage.

Fix the two file-level comments and this is ready.

yagudaev:
Addressed both BLOCKERs in the latest commit: removed the file-level descriptive comments at the top of `desktop/src/renderer/src/lib/voice-prefs.ts` and `desktop/src/main/voice-prefs.ts`. Both files now start directly with imports, matching the project comment policy.

The non-blocking `updateModel` race in `SettingsPage.tsx:207–211` is acknowledged — agreed it's a visual glitch only since `ChatPage` re-reads at call start and `isVoiceForProvider` filters cross-catalog values. Leaving as-is for this PR.

PR #306, opened 2026-04-30
feat(desktop): typed text input in chat composer (NAN-638)

yagudaev:
## Why

Voice-only is a poor fit for sharing exact strings — links, file paths, code snippets, file system paths with non-ASCII characters. Transcription mangles URLs, drops query params, and renames slashes. This PR adds a persistent typed-text composer at the bottom of the desktop chat window so users can drop in precise text and let the assistant respond, both during a call and outside one.

Linear: NAN-638

## What

- **`<ChatComposer>`** at the bottom of `ChatPage` — single-row textarea that grows 1–6 rows, send button on the right, persistent (no hidden-behind-button mode).
- **Enter sends, Shift+Enter inserts newline.** No modifier keys send by accident.
- **Trim-trailing-only normalization** (`normalizeComposerText`): preserves leading whitespace and internal indentation so pasted code, file paths, and query strings survive intact.
- **Empty / whitespace-only is a no-op.** Send button disables until input is meaningful.
- **In-call delivery via `useRealtime.sendUserText(text)`** — emits the existing relay `text.input` event. Both Gemini and OpenAI adapters already accept `injectContext` / `input_text` so the model treats it as a normal user turn.
- **Out-of-call delivery via `streamTextChat`** — opens a one-shot WebSocket to the relay (mirrors the mobile `streamRealtimeText` pattern), sends `session.config` + `text.input`, streams the assistant reply, persists both messages to local SQLite. Same brain agent + tools as voice.
- **Subtle "typed" indicator** on user bubbles created via the composer (small keyboard glyph + "typed" label). Voice messages are unmarked. Indicator is session-scoped (no DB schema change).
- **Vitest coverage** for the pure helpers (`composer.test.ts`, 13 cases).
- **Starlight docs** at `docs/src/content/docs/desktop/typing-messages.mdx`, added to the sidebar in `docs/astro.config.mjs`.

## Key decisions

- **Text user messages are persisted client-side at submit time, not via `transcript.done`.** The relay forwards `text.input` to the model but does not echo a user-side transcript. Persisting at submit makes the bubble appear immediately and survives if the network drops before the assistant replies.
- **Out-of-call uses a fresh one-shot relay session, not a separate `ask_brain` IPC.** There's no existing `ask_brain` IPC handler in `desktop/src/main/`; the relay's text-input pipeline already runs through the brain agent (and tavily web_search). Reusing it keeps text and voice on the exact same backend.
- **No DB migration.** "Typed" cue is stored in component state (`Set<number>` of message IDs from this session). If we ever need it cross-restart we can add a `kind` column later.
- **No semicolons; helpers at file bottom; no ticket IDs in source comments** per the project's `CLAUDE.md`.

## Test plan

- [x] `yarn workspace voiceclaw-desktop typecheck` — passes
- [x] `yarn workspace voiceclaw-desktop test src/renderer/src/lib/composer.test.ts` — 13/13 pass
- [x] `yarn workspace voiceclaw-desktop build` — passes (electron-vite, all 3 entry points)
- [x] `yarn workspace voiceclaw-docs build` — passes; `/desktop/typing-messages/` page generated
- [ ] Manual: launch desktop, type "hello" outside a call → assistant replies in chat
- [ ] Manual: start a call, type a URL with query string → assistant references the exact URL in its spoken reply
- [ ] Manual: paste indented code snippet → indentation preserved in user bubble
- [ ] Manual: press Enter on empty composer → no-op, send button disabled
- [ ] Manual: press Shift+Enter → newline inserted, composer grows to 2 rows

## Pre-existing test failure (not from this PR)

`src/main/services/diagnostic-bundle.test.ts` fails because the `archiver` package isn't in `desktop/package.json`. That file was added before this branch and the failure reproduces on `origin/main`.

### DIFF
--- a/desktop/src/renderer/src/components/ChatComposer.tsx
+++ b/desktop/src/renderer/src/components/ChatComposer.tsx
@@ -1,0 +1,110 @@
+import {
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+  useState,
+  type KeyboardEvent,
+} from 'react'
+import { Send } from 'lucide-react'
+import { Button } from './ui/Button'
+import {
+  COMPOSER_LIMITS,
+  isSubmittable,
+  normalizeComposerText,
+} from '../lib/composer'
+
+export interface ChatComposerProps {
+  onSubmit: (text: string) => void
+  disabled?: boolean
+  placeholder?: string
+}
+
+export function ChatComposer({ onSubmit, disabled, placeholder }: ChatComposerProps) {
+  const [value, setValue] = useState('')
+  const textareaRef = useRef<HTMLTextAreaElement>(null)
+
+  const submit = useCallback(() => {
+    const cleaned = normalizeComposerText(value)
+    if (!cleaned) return
+    onSubmit(cleaned)
+    setValue('')
+  }, [value, onSubmit])
+
+  const handleKeyDown = useCallback(
+    (e: KeyboardEvent<HTMLTextAreaElement>) => {
+      if (e.key === 'Enter' && !e.shiftKey && !e.metaKey && !e.ctrlKey && !e.altKey) {
+        e.preventDefault()
+        submit()
+      }
+    },
+    [submit],
+  )
+
+  useLayoutEffect(() => {
+    const el = textareaRef.current
+    if (!el) return
+    autosize(el)
+  }, [value])
+
+  useEffect(() => {
+    const el = textareaRef.current
+    if (!el) return
+    autosize(el)
+  }, [])
+
+  const canSubmit = !disabled && isSubmittable(value)
+
+  return (
+    <div className="px-4 py-3 border-t border-border bg-background/80 backdrop-blur">
+      <div className="flex items-end gap-2">
+        <textarea
+          ref={textareaRef}
+          value={value}
+          onChange={(e) => setValue(e.target.value)}
+          onKeyDown={handleKeyDown}
+          rows={1}
+          disabled={disabled}
+          placeholder={placeholder ?? 'Type a message — Enter sends, Shift+Enter for newline'}
+          aria-label="Type a message"
+          className="
+            flex-1 resize-none rounded-md border border-input bg-background px-3 py-2
+            text-sm text-foreground placeholder:text-muted-foreground leading-5
+            focus:outline-none focus:ring-2 focus:ring-primary/50 focus:border-primary/50
+            disabled:opacity-50 disabled:cursor-not-allowed
+            whitespace-pre-wrap
+          "
+          style={{ minHeight: minHeightPx(), maxHeight: maxHeightPx() }}
+        />
+        <Button
+          variant="default"
+          size="icon"
+          onClick={submit}
+          disabled={!canSubmit}
+          aria-label="Send message"
+          title="Send (Enter)"
+        >
+          <Send size={18} />
+        </Button>
+      </div>
+    </div>
+  )
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function minHeightPx() {
+  return COMPOSER_LIMITS.minRows * COMPOSER_LIMITS.lineHeightPx + COMPOSER_LIMITS.verticalPaddingPx
+}
+
+function maxHeightPx() {
+  return COMPOSER_LIMITS.maxRows * COMPOSER_LIMITS.lineHeightPx + COMPOSER_LIMITS.verticalPaddingPx
+}
+
+function autosize(el: HTMLTextAreaElement) {
+  el.style.height = 'auto'
+  const next = Math.min(el.scrollHeight, maxHeightPx())
+  el.style.height = `${Math.max(minHeightPx(), next)}px`
+}
--- a/desktop/src/renderer/src/components/ChatComposer.tsx
+++ b/desktop/src/renderer/src/components/ChatComposer.tsx
@@ -34,6 +34,7 @@ export function ChatComposer({ onSubmit, disabled, placeholder }: ChatComposerPr
   const handleKeyDown = useCallback(
     (e: KeyboardEvent<HTMLTextAreaElement>) => {
       if (e.key === 'Enter' && !e.shiftKey && !e.metaKey && !e.ctrlKey && !e.altKey) {
+        if (e.nativeEvent.isComposing) return
         e.preventDefault()
         submit()
       }
--- a/desktop/src/renderer/src/lib/composer.ts
+++ b/desktop/src/renderer/src/lib/composer.ts
@@ -1,0 +1,30 @@
+// Pure helpers for the chat composer. Kept separate from the React
+// component so they're easy to unit-test.
+
+const MIN_ROWS = 1
+const MAX_ROWS = 6
+const LINE_HEIGHT_PX = 20
+const VERTICAL_PADDING_PX = 16
+
+export function normalizeComposerText(value: string): string {
+  // Trim trailing whitespace only — preserve leading and internal whitespace
+  // because pasted code, indented diffs, and file paths depend on it.
+  return value.replace(/[ \t\r\n]+$/u, '')
+}
+
+export function isSubmittable(value: string): boolean {
+  return normalizeComposerText(value).length > 0
+}
+
+export function computeTextareaHeight(value: string): number {
+  const lines = Math.max(MIN_ROWS, value.split('\n').length)
+  const clamped = Math.min(MAX_ROWS, lines)
+  return clamped * LINE_HEIGHT_PX + VERTICAL_PADDING_PX
+}
+
+export const COMPOSER_LIMITS = {
+  minRows: MIN_ROWS,
+  maxRows: MAX_ROWS,
+  lineHeightPx: LINE_HEIGHT_PX,
+  verticalPaddingPx: VERTICAL_PADDING_PX,
+}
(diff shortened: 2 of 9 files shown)


desktop/src/renderer/src/lib/composer.ts
```diff
@@ -0,0 +1,30 @@
+// Pure helpers for the chat composer. Kept separate from the React
```
yagudaev: **BLOCKER — file-level comment reads like a commit-message paragraph**

`composer.ts` opens with:
```
// Pure helpers for the chat composer. Kept separate from the React
// component so they're easy to unit-test.
```

Per CLAUDE.md: *"Never reference tickets, bugs, or fix history in source comments. No NAN-XXX, no 'fixes the doubled-words bug', no 'this used to do X'. That goes in the commit body and PR description — it rots in source."* and *"Don't restate what the code does. Well-named identifiers do that."* — reviewers must treat this as a BLOCKER.

The filename `composer.ts` and the `export function normalizeComposerText` / `isSubmittable` names already communicate the purpose. Delete this header comment entirely.

desktop/src/renderer/src/lib/text-chat.ts
```diff
@@ -0,0 +1,111 @@
+// One-shot text turn over the relay WebSocket. Used when the chat
```
yagudaev: **BLOCKER — file-level commit-message comment**

`text-chat.ts` opens with a 4-line paragraph:
```
// One-shot text turn over the relay WebSocket. Used when the chat
// composer submits while no realtime call is active — opens a fresh ws,
// sends session.config + text.input, streams the assistant transcript,
// closes when the relay reports the turn is done.
```

This restates what the code does, names internal event types (`session.config`, `text.input`), and reads like a PR description pasted into source. CLAUDE.md rule: *"Don't restate what the code does. Well-named identifiers do that."* — treat as BLOCKER same as PR #300.

Delete all four lines. The export name `streamTextChat` and the `TextChatOptions` / `TextChatCallbacks` interfaces are self-describing.

desktop/src/renderer/src/components/ChatComposer.tsx
```diff
@@ -0,0 +1,110 @@
...
+  type KeyboardEvent,
+} from 'react'
+import { Send } from 'lucide-react'
+import { Button } from './ui/Button'
+import {
+  COMPOSER_LIMITS,
+  isSubmittable,
+  normalizeComposerText,
+} from '../lib/composer'
+
```
yagudaev, line 16: **BLOCKER — IME composition not guarded; Enter fires during CJK composition**

`handleKeyDown` checks modifier keys but does not check `e.nativeEvent.isComposing`. On macOS with Japanese/Chinese/Korean input, pressing Enter to confirm a composition candidate will trigger `submit()` and send the in-progress composition as a user message. Standard fix:

```tsx
if (e.key === 'Enter' && !e.shiftKey && !e.metaKey && !e.ctrlKey && !e.altKey) {
  if ((e.nativeEvent as InputEvent).isComposing) return  // let IME finish
  e.preventDefault()
  submit()
}
```

`nativeEvent.isComposing` is `true` from `compositionstart` through `compositionend`. Without this guard every Enter keypress during CJK input submits the raw unconfirmed text.

desktop/src/renderer/src/pages/ChatPage.tsx
```diff
@@ -402,12 +405,99 @@ export function ChatPage() {
...
+        setIsThinking(false)
+        setStreamingText(full)
+      },
+      onDone: async (full) => {
+        setStreamingText('')
+        setIsThinking(false)
+        if (full.trim()) {
+          await addMessage(convId, 'assistant', full)
+          await loadMessages()
+        }
```
yagudaev, line 486: **BLOCKER — parallel submit leak: cancel return value is dropped**

`streamTextChat` returns a cancel function, but `handleComposerSubmit` calls it as a fire-and-forget expression:

```ts
streamTextChat(text, { ... }, { ... })
```

The cancel function is discarded. If the user submits a second message before the first reply completes, a second `streamTextChat` opens a new WebSocket while the first is still live. Both callbacks race to call `setStreamingText` and `addMessage` against the same `convId`. The first session is never cancelled.

Additionally, `ChatComposer` is rendered without a `disabled` prop, so the send button and Enter remain active during streaming. The docs claim *"The composer can disable itself temporarily — for example while a previous typed message is still streaming a reply"* but the code does not pass `disabled={isThinking || !!streamingText}` to `<ChatComposer>`.

Fix: store the cancel function in a ref, call it before starting a new session, and pass `disabled={isThinking || !!streamingText}` to `<ChatComposer>`.

docs/src/content/docs/desktop/typing-messages.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+## During a call vs outside one
+
+The composer works in both states. The path your message takes is different, but the result feels the same.
+
+**During a realtime call.** The typed text is delivered to the live session as a text input alongside whatever you're saying. The assistant treats it as a user turn — the realtime model sees the text, can call tools (including `ask_brain`), and responds with voice and a transcript exactly like a spoken turn. You'll hear the answer if your speakers are on, and it lands in the chat history. Useful when you want to drop the assistant a URL mid-conversation without breaking the flow.
+
+**Outside a call.** Typing a message opens a one-shot relay session, sends your text, streams the assistant's reply into the chat as text, and closes. No microphone access, no speaker output — pure text-in / text-out. Your message and the reply are both saved to the conversation, so the next time you start a call the model has the full context.
+
```
yagudaev, line 37: **Docs audience leak — internal event names in user-facing page (HARD rule)**

Lines 34 and 36 name internal implementation details:

- Line 34: `ask_brain` — this is a tool name visible only in the relay/brain layer
- Line 36: *"opens a one-shot relay session"* / *"relay session"* — `relay` is infrastructure vocabulary; users only know they're chatting

Per `feedback_docs_audience.md`: *"implementer specs go in PR bodies / backlog tasks, not the docs site (HARD RULE)"*.

Rewrite line 34 without `ask_brain`: *"The assistant treats it as a normal turn and can call tools — including web search"*.

Rewrite line 36 without `relay`: *"Typing a message connects to your assistant in the background, streams the reply into the chat as text, then disconnects."*

yagudaev (commented):
## Review verdict: request-changes (4 BLOCKERs)

Typecheck passes. 13/13 tests green. Architecture (one-shot WS, no new IPC, session-scoped typed-ID tracking) is sound. The in-call `sendUserText` path is clean. Most of the code is in good shape — four issues need a fix before merge.

---

### BLOCKERs

**1. `composer.ts` line 1-2 — file-level commit-message comment**
Delete the two-line header. `composer.ts` + the export names communicate the purpose. CLAUDE.md: reject comments that restate what the code does. Same pattern was rejected in PR #300.

**2. `text-chat.ts` line 1-4 — file-level commit-message comment with internals in it**
Delete all four lines. Exposes `session.config` / `text.input` event names in source, reads like a PR description. Same rule.

**3. `ChatComposer.tsx` — Enter fires during IME/CJK composition**
`handleKeyDown` does not check `e.nativeEvent.isComposing`. On macOS with Japanese/Chinese/Korean input, pressing Enter to confirm a composition candidate submits immediately. Add `if ((e.nativeEvent as InputEvent).isComposing) return` before `e.preventDefault()`.

**4. `ChatPage.tsx` — `streamTextChat` cancel value dropped; composer not disabled during streaming**
The cancel function returned by `streamTextChat` is thrown away. A second submit while the first reply is streaming opens a parallel WS with racing `addMessage` / `setStreamingText` calls. Store the cancel fn in a ref and call it at the top of `handleComposerSubmit`. Also: `<ChatComposer>` is rendered without `disabled={isThinking || !!streamingText}` — docs say the composer disables itself during streaming but the code does not enforce it.

---

### Docs audience (HARD rule) — partial fail

`typing-messages.mdx` is mostly well-written but leaks two implementer terms:
- Line 34: `` `ask_brain` `` — internal tool name, not user vocabulary
- Line 36: "one-shot relay session" / "relay session" — relay is infrastructure, invisible to users

Both need user-facing rewrites. See inline comment.

---

### Code style

No ticket IDs (NAN-638) found in source files — good. Semicolons absent. Helpers at bottom of files. The file-level comments are the only style violations and those are already flagged as BLOCKERs above.

---

### What's good

- Whitespace normalization (`/[ \t\r\n]+$/u`) is correct and the 7 test cases cover it well.
- `isSubmittable` correctly delegates to `normalizeComposerText`, so button state and submit guard are consistent.
- `finish()` guard in `text-chat.ts` (`if (didFinish) return`) correctly handles the race between `onerror` and `onclose`.
- `typedMessageIds` as session-scoped `Set<number>` is the right call — no DB migration needed.
- Brand voice in docs is warm and editorial throughout, no generic SaaS/cyan language.
- `transcript.done` → `turn.ended` fallback in the WS handler is a solid defensive pattern.

Fix the four BLOCKERs and this is ready.

yagudaev:
Addressed all 4 BLOCKERs + the docs audience violation in the latest commit.

1. **`composer.ts` / `text-chat.ts` file-level comments** — deleted; both files start at imports.
2. **IME composition guard** — `ChatComposer.tsx:36` now `if (e.nativeEvent.isComposing) return` before `e.preventDefault()`.
3. **Cancel inflight + disable during streaming** — `ChatPage.tsx` captures `streamTextChat`'s cancel fn in `textChatCancelRef`, cancels any prior call before submitting a new one, clears on `onDone`/`onError`, and on unmount. Composer now receives `disabled={isThinking || !!streamingText}`.
4. **Docs audience** — `typing-messages.mdx` rewritten to drop `ask_brain`, "relay server URL", "one-shot relay session". User-facing only.

Verified: typecheck green, 70/70 desktop tests pass, build green. (Pre-existing `archiver`-missing failure on `diagnostic-bundle.test.ts` is unchanged on `origin/main`, called out in the PR body.)

PR #307, opened 2026-04-30
spike: tool-issues investigation (NAN-677)

yagudaev:
# Summary

Diagnostic spike for NAN-677. **No code changes** — read-only investigation across `relay-server/`, `desktop/`, the in-house tracer DB (~/.voiceclaw/tracing.db), and recent ticket history. Empty commit so the PR can carry the findings.

Scope: provider differences (Gemini / OpenAI / xAI Grok), tool-result delivery, error paths, latency, NAN-690's UI rows, ask_brain reliability, multi-tool chains, malformed-args paths.

## Top 3 findings (ranked by user impact)

1. **~50% of ask_brain calls returned empty over the last week** (63/125 trace observations across 2026-04-23..2026-04-28; the run on 2026-04-29 was clean — likely after a model-provider swap per `docs/operations/brain-empty-response.mdx`). The known fix is operator-side (swap `claude-cli` → `openai-codex`), but we ship without that workaround in the bundled gateway and have no in-product self-heal or retry. Severity P0.
2. **Tool-call rows leak across conversations and don't persist.** `setToolCalls([])` only fires in `newConversation`; `loadConversation`/`loadLatestConversation` (ChatPage.tsx:144-161) leave stale tool entries on screen. Tool calls are also held in React state only — never written to SQLite — so they vanish on app restart and on conversation switch from a fresh launch the inline rows shipped in NAN-690 simply are not there. Severity P1.
3. **`ask_brain` blocks for 120s on backend stalls** (multiple traces hit the exact 120,000 ms ceiling). On Gemini's async path the user hears the "let me check" filler then complete silence for two minutes; on OpenAI/xAI the model is stuck waiting on `function_call_output`. There is no soft timeout, no streaming "still thinking" prompt past the SSE step events, and no abort surfaced to the user. Severity P1.

### F1. Brain agent: empty-response rate ~50% (P0)
**Where:** `relay-server/src/tools/brain.ts:158`, `docs/src/content/docs/operations/brain-empty-response.mdx`.
**Evidence:** `~/.voiceclaw/tracing.db`:
```
sqlite> SELECT COUNT(*) total,
        SUM(CASE WHEN attributes_json LIKE '%Empty response%' THEN 1 ELSE 0 END) empty
        FROM observations WHERE name='ask_brain'
        AND start_time_ns > strftime('%s','now','-30 days')*1000000000;
125|63
```
Per-day breakdown: 2026-04-25=16/22 empty, 04-27=16/21, 04-28=18/34; 04-29=0/19 (after the operator workaround landed).
**Root cause hypothesis:** OpenClaw gateway was using `claude-cli/...` as `agents.defaults.model.primary`, which Anthropic intermittently blocks for automation. The relay sees a stream that closes with no `data:` chunks → `fullResponse=""` → returns `{"error":"Empty response from brain agent"}`. The error string is then JSON-encoded and shoved back to the model as if it were a normal reply (mixed wire format — see F4).
**Proposed fix:** Combination of (a) ship a default OpenClaw config that does not point at `claude-cli` for new installs, (b) add retry-with-fallback in the relay so a single empty SSE doesn't surface to the user, and (c) include captured stderr in the gateway's failure response so the relay log isn't blind.

### F2. Tool-call rows vanish across conversation switches (P1)
**Where:** `desktop/src/renderer/src/pages/ChatPage.tsx:144-161, 405-412`.
- `loadConversation(id)` and `loadLatestConversation()` overwrite `messages` but never touch `toolCalls`.
- `setToolCalls([])` only runs from `newConversation()`.
- `toolCalls` is `useState<ToolCallEntry[]>` — never persisted to `~/Library/.../voiceclaw.db`.
**Impact:** Switching from conversation A → B leaves A's "ask_brain · done · 12.4s" rows interleaved into B's timeline (sorted purely by timestamp). After app restart, every prior conversation's tool calls are gone — the inline rows shipped in NAN-690 are session-scoped, not conversation-scoped.
**Proposed fix:** Add a `tool_calls` table (`call_id PK`, `conversation_id`, `name`, `args`, `result`, `error`, `status`, `started_at`, `duration_ms`); persist on `onToolCallStarted/Completed/Failed`; load alongside messages in `loadConversation`. Clear React state in `loadConversation` before re-hydrating.

### F3. Brain async path can stall for 120s with no user feedback (P1)
**Where:** `relay-server/src/tools/brain.ts:26`, `relay-server/src/session.ts:350-446`.
- Local timeout set to 120 s in `askBrain` (`local 120s timeout`). The trace DB shows multiple calls hitting exactly 120009 ms / 120006 ms / 114149 ms before aborting.
- During the wait, `tool.progress` SSE chunks (`step_complete`) are forwarded to the client and replace `streamingText` (ChatPage.tsx:233-237), but if openclaw stalls before emitting any step event, the client just sits on "Looking into it now…" until the 120 s ceiling.
- After abort, the model never gets a `function_call_output` (Gemini) or sees an injected error context (OpenAI/xAI gets `injectContext("[Brain agent failed for query…]")`). For the **blocking** ask_brain branch the abort path DOES feed a cancellation result, but the **async** path (default for Gemini) does not call `injectContext` on abort — see `session.ts:413-418`: only the cancel branch handles it; the abort-while-async case relies on `controller.signal.aborted` which only fires from upstream cancel events, not from the local 120 s timeout. The local timeout aborts the controller in `brain.ts:26` directly, so when the catch block in session.ts runs, `controller.signal.aborted` is false on the **session-level** controller (different controller object) and we DO inject the failure text — but the model is, by that point, two minutes into a dead silence and the user has hung up.
**Proposed fix:** (a) Lower the brain hard-timeout to 30-45 s and surface a structured timeout error earlier; (b) emit a synthetic `tool.progress` "still working" beat every ~10 s while the SSE is silent; (c) in async mode, inject a "the brain took too long, want me to try again?" prompt on timeout instead of "Let the user know the search didn't work".

### F4. Brain error payload format is inconsistent with success format (P2)
**Where:** `relay-server/src/tools/brain.ts:89, 95, 113, 158`.
- Success path returns plain text (`return fullResponse`).
- Error path returns `JSON.stringify({error: "..."})`.
- Both feed the model via the same `injectContext`/`sendToolResult`. Gemini's `sendToolResult` parses the JSON and wraps as `{response: parsed}`, but OpenAI's path passes raw output as `function_call_output.output` — so error vs success carries different shapes for the same `function_call_output` slot. The model sometimes parrots the literal error string ("Empty response from brain agent") to the user.
**Proposed fix:** Always return JSON-encoded `{ok: true, text: ...}` or `{ok: false, error: ...}` from `askBrain`, and have the adapters / blocking dispatcher unwrap consistently. Update prompt instructions to handle the unified shape.

### F5. `executeSyncTool` JSON.parse is uncaught (P2 — defensive)
**Where:** `relay-server/src/tools/index.ts:144`.
```ts
case "echo_tool": {
  const parsed = JSON.parse(args)   // throws on malformed args
  return JSON.stringify({ echoed: parsed.message })
}
```
The session caller (`session.ts:156-162`) doesn't wrap this in try/catch. A model that emits malformed JSON for `echo_tool` args would crash `handleServerToolCall` and leave `pendingToolCalls` desynced. Low impact since `echo_tool` is a test tool, but worth a 1-line `try { ... } catch { return JSON.stringify({error: "invalid arguments"}) }`.

> **Trivial fix sketch (do not apply, included so the user can decide):**
> ```ts
> case "echo_tool": {
>   try {
>     const parsed = JSON.parse(args)
>     return JSON.stringify({ echoed: parsed.message })
>   } catch (e) {
>     return JSON.stringify({ error: e instanceof Error ? e.message : "invalid arguments" })
>   }
> }
> ```

### F6. `displayText` tool referenced but never registered (P2)
**Where:** `desktop/src/renderer/src/pages/ChatPage.tsx:213-222`.
The renderer special-cases `name === 'displayText'` to write the tool's `args.text` into the messages table. But `displayText` is **not** in `getRelayTools` / `executeSyncTool` (`relay-server/src/tools/index.ts`). It cannot fire today.
**Hypothesis:** Dead code from an earlier prototype, or a planned tool that never landed. Either implement it server-side (as a sync tool that returns `{ok: true}` and lets the client persist the text) or delete the dead branch.

### F7. `ToolCallRow` "Show result" hidden when result is empty (P2)
**Where:** `desktop/src/renderer/src/components/ToolCallRow.tsx:65-77`.
The "Show result" toggle is gated on `result` being truthy. For `tool.progress` events we update `streamingText` only, never the entry — so a long-running tool with multiple progress steps has nothing to show in its row. Also when status is `error` and `error` is set but `result` is empty, the result chevron is hidden but error text is shown inline (good) — except the `resultExpanded` block on line 85 will never render because the gate above (line 65) hid the expander. Two interacting branches → confusing for errors with stack-trace-like detail.
**Proposed fix:** Track per-entry `progressSteps: string[]` on the store; render most recent step inline while in-progress; keep the expander available even without a final `result` so users can see the progress trail.

### F8. Tool-call deltas: no provider streams partial args (informational)
The ticket asks about "tool-call deltas". None of the three providers expose this:
- **OpenAI/xAI** emit `response.function_call_arguments.delta` events, but `openai.ts:513` explicitly suppresses logging them and the adapter only fires `tool.call` on `.done`.
- **Gemini** sends `toolCall.functionCalls` with full args, no delta protocol.
**Implication:** UI cannot show "tool args being typed". This is fine for echo/web_search but means a slow-thinking model emitting a 4 KB ask_brain query has zero visible progress until it commits. Not a bug, but worth noting.

### F9. Multi-tool: parallel calls work, sequential chains untested (P2)
- Both adapters increment `pendingToolCalls` on each call and decrement on each `sendToolResult`. `requestResponse` only fires when no response is active, so two parallel tools work correctly (cancel + queue + flush on `response.done`).
- However, **a tool calling another tool** (`ask_brain` → openclaw decides to invoke `web_search` → results back) is NOT a thing in the current architecture. The relay only routes calls the realtime model emits; openclaw's internal tool use is hidden inside the gateway's response. So "multi-tool chains" exist only inside openclaw, and are opaque to the relay's tracing/UI.
**Proposed fix:** If we want to surface openclaw's nested tool calls, extend the SSE protocol so openclaw emits typed `tool_started` / `tool_finished` step events alongside `step_complete`, and have the relay forward them as nested rows.

### F10. xAI Grok Voice tool-result wire format inherited untested (P2)
**Where:** `relay-server/src/adapters/xai.ts` (extends `OpenAIAdapter`).
The xAI adapter sends `function_call_output` items the same way OpenAI does. The xAI Realtime docs are scant on whether the same `conversation.item.create` / `function_call_output` shape is supported, or whether it expects a different envelope. Tracer DB shows zero xAI tool-call observations in the last 30 days (Gemini-dominant traffic), so this is unverified in production.
**Proposed fix:** Add an integration smoke test against xAI that drives `echo_tool` end-to-end. NAN-655 added the model but didn't add a tool-call test.

### F11. No persisted tool-call audit trail in local DB (P1, overlap with F2)
Even ignoring conversation-switching, today there is no way for a user to look at `~/Library/Application Support/voiceclaw/voiceclaw.db` and see what tools ran, with what args, when. The OTel tracer captures it (`~/.voiceclaw/tracing.db`), but that's developer-only. The diagnostic-bundle export (PR #293) doesn't include the tracer DB, so a user reporting "the brain said nothing" can't ship us the tool-call evidence.
**Proposed fix:** Tied to F2 — the same `tool_calls` table doubles as user-facing audit + diagnostic-bundle attachment.

### F12. Test coverage gaps (informational)
**Coverage we have:**
- `relay-server/test/session/tool-blocking.test.ts` — blocking vs non-blocking dispatch, capability fallback, abort cancellation.
- `relay-server/test/brain/brain-result-event.test.ts` — `brain.result` event emission on success/error/cancel.
- `relay-server/test/brain/traceparent.test.ts` — W3C trace context injection.

**Gaps:**
- No malformed-JSON args test for any tool.
- No empty-SSE-response test for `askBrain` (the main reliability bug).
- No tool-call test for the OpenAI or xAI adapters at all (only Gemini reconnect / unit tests).
- No desktop-renderer test for the `tool-call-store`, `ToolCallRow`, or the timeline merge in `ChatPage`.
- No test for parallel/multi-tool dispatch.

## Root cause hypotheses

- **Brain emptiness (F1):** OpenClaw model backend (`claude-cli`) is the proximate cause; the relay's "if empty, treat as error" treats a symptom not the disease. Operator-side fix is documented but not packaged.
- **UI tool-call leakage (F2):** NAN-690 shipped tool rows as ephemeral state to get the inline UI in fast; persistence was deferred. The conversation-switch leak is an unintended consequence of putting `toolCalls` at the same scope as the realtime hook callbacks.
- **Latency cliff (F3):** `askBrain` was tuned for "exec approval may need 2 min" — sensible for a developer-driven workflow, hostile for voice. The async path's progress signaling depends on openclaw emitting step events, which it doesn't always do.
- **Inconsistent error payloads (F4):** Each error site in `brain.ts` writes its own ad-hoc `JSON.stringify({error: ...})`; never refactored after the SSE streaming path landed.

## Proposed follow-up tickets

Listed for the user to triage and create as needed.

1. **NAN-???: Persist tool calls per-conversation in local SQLite + clear stale rows on conversation switch (P1)** — Add `tool_calls` table; persist on every state transition; load alongside messages; clear React state in `loadConversation`. Includes the diagnostic-bundle attachment hookup. (Covers F2 + F11.)

2. **NAN-???: Eliminate ask_brain empty-response failure mode in the bundled stack (P0)** — Default new installs to a non-`claude-cli` primary model; add a single retry-with-fallback in `askBrain` when a stream completes with zero deltas; surface gateway stderr in the failure payload. (Covers F1.)

3. **NAN-???: Brain-call user-feedback during long stalls + lower hard timeout (P1)** — Drop `askBrain` timeout from 120 s to 30-45 s; emit synthetic "still working" `tool.progress` events every ~10 s of SSE silence; on timeout, inject a friendly "want me to retry?" prompt instead of "search didn't work". (Covers F3.)

4. **NAN-???: Unify brain tool-result payload shape and update prompt to match (P2)** — All `askBrain` returns become `{ok, text?, error?}`; adapters unwrap consistently; instructions updated so the model never parrots the literal error string. (Covers F4.)

5. **NAN-???: ToolCallRow shows progress steps mid-flight + always-available expander (P2)** — Extend `tool-call-store` with `progressSteps`; update on `tool.progress`; render the most recent step inline while `in-progress`; keep the expander available even when only error is present. (Covers F7.)

6. **NAN-???: xAI Grok tool-call smoke test + delete dead displayText branch (P2)** — Add an integration test that drives `echo_tool` against xAI to verify the inherited OpenAI wire format works there; remove or implement `displayText` in ChatPage. (Covers F6 + F10.)

7. **NAN-???: Defensive try/catch around executeSyncTool + malformed-args test (P2 — quick win)** — Wrap the `JSON.parse` in `executeSyncTool`; add a malformed-args test for each tool; ensures a misbehaving model can't desync `pendingToolCalls`. (Covers F5.)

8. **NAN-???: Tool-call test coverage backfill (P2)** — Add tests for: malformed args (all tools), empty-SSE response (askBrain), OpenAI adapter tool-call dispatch, parallel multi-tool, ToolCallRow state transitions. (Covers F12.)

## Test plan

- [x] Read `desktop/src/renderer/src/lib/use-realtime.ts`, `pages/ChatPage.tsx`, `lib/tool-call-store.ts`, `components/ToolCallRow.tsx`.
- [x] Read all `relay-server/src/adapters/*.ts` and `relay-server/src/tools/*.ts` plus `session.ts`.
- [x] Spot-check `~/.voiceclaw/tracing.db` for ask_brain / web_search outcomes (last 30 days).
- [x] Cross-reference `docs/src/content/docs/operations/brain-empty-response.mdx`.
- [x] Survey existing tool-related tests (`relay-server/test/session`, `relay-server/test/brain`, `relay-server/test/adapters`).
- [ ] No live tool-call repro (out of scope — autonomous).


yagudaev:
Spike complete — findings live in this PR body. Closing without merge since there's no code to ship; follow-up implementation tickets to be triaged from the proposed list in the description.

PR #309, opened 2026-04-30
feat(desktop): image attachments in chat (NAN-646 MVP)

yagudaev:
## Why

Conversations in the desktop app are voice-only today, but a lot of real questions are easier with a picture: a screenshot of an error, a photo of a whiteboard, a sketch. NAN-646 ("attach files to desktop conversations") asks for the broader story; this PR ships the smallest useful slice — **images only** — so we can prove the multimodal pipeline end-to-end before widening it to PDFs and other file types.

## What

MVP scope, deliberately narrow:

- Three ways to attach an image: **drag & drop** onto the chat area, **paste** from clipboard (`Cmd+V`), or **Attach** button → native file picker (Electron `dialog.showOpenDialog`).
- Allowed types: `image/png`, `image/jpeg`, `image/webp`. Max size 10 MB. Validation runs both renderer-side (before queueing) and main-side (before persisting), with a friendly error row above the tray for rejects.
- Pending tray with thumbnails + per-image remove + Send. Send creates a single new user message with a short placeholder body (`Attached: photo.png`) and attaches every pending image to it.
- Storage decision: **dedicated `message_attachments` table** with `ON DELETE CASCADE`. Files ≤ 5 MB stored inline as base64; larger files written to `app.getPath('userData')/attachments/<uuid>.ext` and referenced by path. Disk-spill keeps the SQLite DB lean while still surviving DB delete via the same code path.
- Inline render in the user `MessageBubble` (click thumbnail → opens full-size in a new window).
- **Brain integration:** when a call is active, attached images are pushed via the existing `frame.append` channel that screen-share already uses, so the realtime model sees them as live visual context. Out-of-call attachments persist in chat history but are not retroactively replayed when a new call starts (documented as an MVP limitation; multimodal `text.input` parts on the relay would be a follow-up).
- Cascade safety: `db:deleteMessage`, `db:deleteConversation`, and `db:deleteAllConversations` all collect file-storage paths *before* DB rows go away and unlink them on disk — no orphan files, no surprise data loss outside the explicit delete the user asked for.

## Out of scope (explicit follow-ups)

- PDFs, plain-text code, generic documents.
- Image compression/re-encoding.
- Per-attachment delete (attachments are tied to their parent message).
- Multimodal injection of attachments into a *new* call's context.

## Screenshots

User has not provided UI reference screenshots yet — the implementation matches the spec in the ticket. Reviewer: please request a screenshot of the drag overlay, the pending tray, and an inline-rendered image bubble before approving the visual treatment.

## Test plan

- [x] `yarn typecheck` clean (desktop)
- [x] `yarn test` — 71 passed (13 new vitest cases for attachment validation, mime whitelist, size cap, inline-vs-disk threshold). Pre-existing `diagnostic-bundle.test.ts` failure due to missing `archiver` module is **unrelated** and reproduces on `origin/main`.
- [x] `yarn build` (desktop) — clean
- [x] `yarn build` (docs) — Starlight builds the new `desktop/attachments` page
- [ ] Manual: drag-drop a PNG, JPG, WEBP onto chat → renders inline, persists across reload
- [ ] Manual: paste `Cmd+Shift+4` screenshot → lands in pending tray, sends successfully
- [ ] Manual: Attach button → native picker opens with PNG/JPG/WEBP filter
- [ ] Manual: 12 MB PNG → rejected with clear error
- [ ] Manual: GIF / PDF dropped → rejected with mime error
- [ ] Manual: send during active call → assistant references the image
- [ ] Manual: delete message → attachment file removed from `userData/attachments/`
- [ ] Manual: delete conversation → all attachments removed

## Docs

New page: `docs/src/content/docs/desktop/attachments.mdx`, sidebar entry under "Guides → Desktop: image attachments". Covers UX, validation rules, deletion cascade, and the explicit MVP limitations.


desktop/src/renderer/src/lib/db.ts
```diff
@@ -133,6 +180,39 @@ export async function deleteMessage(
...
+export async function attachToMessage(
+  messageId: number,
+  input: AttachmentInput,
+): Promise<{ ok: true; attachment: Attachment } | { ok: false; error: string }> {
+  return api().attachToMessage(messageId, input)
+}
+
+export async function getAttachmentsForMessage(messageId: number): Promise<Attachment[]> {
+  return api().getAttachmentsForMessage(messageId)
+}
```
yagudaev, line 192: **BLOCKER — `file://` thumbnails will be broken images for 5–10 MB attachments**

`attachmentDataUrl` returns `file://${attachment.path}` for disk-stored attachments. Electron renders the renderer in an `https://` (or custom protocol) origin with `webSecurity: true` (default). Loading a cross-origin `file://` resource from an `` is blocked by the same-origin policy — the thumbnail silently renders as a broken image.

The click-to-open path happens to work because `window.open(fileUrl, "_blank")` hits `setWindowOpenHandler → shell.openExternal`, which hands off to the OS. But the inline thumbnail never loads for any attachment that was spilled to disk (> 5 MB).

**Fix options:**
1. Register a `safe-file://` protocol in main that validates the path is under `app.getPath("userData")/attachments/` and streams the file, then use that URL here.
2. Add an IPC (`attachments:readAsBase64`) that validates the path, reads the file, and returns base64 — reconstruct the `data:` URL on the renderer side.

Option 2 is simpler for MVP; option 1 is cleaner for large files.

docs/src/content/docs/desktop/attachments.mdx
```diff
@@ -0,0 +1,71 @@
...
+- **The Attach button.** Click **Attach** in the top-right of the chat header. A native open-file dialog appears, filtered to PNG / JPG / WEBP.
+
+In all three cases the image lands in a small **pending tray** at the bottom of the chat. You can drop several images in a row before sending; each shows a thumbnail with a small `×` you can click to remove it before it's committed. Click **Send** to attach all pending images to a new user message in the conversation.
+
+## What happens after you click Send
+
+When you send pending attachments:
+
+1. A new user message is added to the conversation with a short placeholder (e.g. `Attached: photo.png`). This is what the conversation list and history will show as the message body.
+2. Each image is persisted alongside the message in the local SQLite database. Images at or under 5 MB are stored inline as base64; larger files (up to the 10 MB ceiling) are written to a file under your Electron `userData/attachments` directory and referenced by path. The behaviour is automatic — you don't pick storage manually.
```
yagudaev: **BLOCKER — Audience violation: implementer details in user-facing docs** (`feedback_docs_audience.md` — hard rule)

Line 27 exposes internal implementation mechanics that a user does not need and should not see:
- "local SQLite database"
- "stored inline as base64"
- "written to a file under your Electron `userData/attachments` directory"
- "5 MB … inline … larger files … referenced by path"

A user reading this page wants to know: "will my images be saved?" and "where do they go when I delete a message?" They do not need to know the storage algorithm, the threshold, or the database technology.

Replace line 27 with something like:

> Each image is saved locally with the message in the conversation history. The behaviour is automatic — you do not need to configure anything.

The inline/disk-spill threshold and `userData/attachments` path belong in the PR body (where they already are), not the docs page.

desktop/src/renderer/src/pages/ChatPage.tsx
```diff
@@ -575,6 +790,29 @@ export function ChatPage() {
...
+            aria-label="Dismiss attachment error"
+          >
+            <Plus size={14} className="rotate-45" />
+          </button>
+        </div>
+      )}
+
+      {/* Pending attachments tray */}
+      <AttachmentTray
+        pending={pendingAttachments}
```
yagudaev, line 912: **Nit (UX gap) — out-of-call attachment send is silently missing the "image not forwarded" notice**

The PR body and docs both call out that images attached while no call is active are **not** forwarded to the model. But the `AttachmentTray` component has no contextual label or tooltip when `!isCallActive`. A user who attaches a photo expecting the assistant to see it during the next call will be surprised.

Consider passing an `isCallActive` prop to `AttachmentTray` and showing a small subtitle such as _"Images will be saved to history. Start a call to share with the assistant."_ This keeps the limitation visible at the point of action, not buried in the docs Limitations section.

yagudaev (commented):
## Review — Request Changes

Two BLOCKERs before merge. Everything else is solid.

---

### BLOCKER 1 — `file://` thumbnails broken for 5–10 MB attachments

**File:** `desktop/src/renderer/src/lib/db.ts`, `attachmentDataUrl`

`attachmentDataUrl` returns `file://${attachment.path}` for disk-stored attachments. Electron's renderer runs with `webSecurity: true` (default, confirmed — no override in `window-lifecycle.ts`). A `` request from an `https://` origin is blocked by the same-origin policy. The thumbnail will silently render as a broken image for every attachment between 5–10 MB.

The `window.open → setWindowOpenHandler → shell.openExternal` path works fine for click-to-open, but the inline display does not.

**Fix:** register a `safe-file://` custom Electron protocol in main (validate path is under `userData/attachments/`) and rewrite the URL, OR add an IPC that reads the file and returns base64 to reconstruct a `data:` URL renderer-side.

---

### BLOCKER 2 — Audience violation in `attachments.mdx`

**File:** `docs/src/content/docs/desktop/attachments.mdx`, line 27

The «What happens after you click Send» step 2 exposes internal implementation details: «local SQLite database», «inline as base64», «Electron `userData/attachments` directory», «5 MB threshold». These are implementer-spec details and violate the `feedback_docs_audience.md` hard rule.

The PR body already documents these correctly. The docs page should just say the image is saved locally — no storage mechanics.

---

### All-clear items

- **Data safety**: unlink logic is scoped to DB-queried rows only, never globs. `deleteAttachmentFile` is best-effort (silent catch), SQL DELETE proceeds regardless. Foreign keys ON, CASCADE in schema.
- **Schema migration**: `CREATE TABLE IF NOT EXISTS` — existing DBs are safe.
- **Code style**: no semicolons, no ticket IDs, no file-level commit-message headers in source. Helpers at bottom of files. Clean.
- **Tests**: 71 passed (13 new). Validation, MIME whitelist, size cap, inline-vs-disk threshold, cascade all covered. The one failing suite (`diagnostic-bundle.test.ts`) is a pre-existing `archiver` module issue, confirmed present on `origin/main`.
- **5 MB boundary**: `shouldStoreInline` uses `<= INLINE_STORAGE_THRESHOLD_BYTES` — exactly 5 MB is stored inline. Tests cover this.
- **Drag/drop**: `dragDepthRef` counter correctly handles child-element enter/leave events. Overlay not stuck.
- **Paste**: only intercepts `item.kind === 'file'` items — text-only pastes pass through unaffected. Mixed text+image pastes: file items are queued, text items are not prevented (no `e.preventDefault()` until files are found). Correct behaviour.
- **frame.append wire format**: `realtime.sendFrame(pending.base64)` matches the existing `{ type: 'frame.append', data: base64Jpeg }` format that screen-share already uses.
- **Path traversal**: disk paths are generated in main (UUID + known extension from a MIME whitelist), never constructed from user input. Low risk.
- **UI screenshots not provided** (noted in PR body) — request screenshots of drag overlay, tray, and rendered bubble before final approval.
- **Out-of-call UX**: Tray has no label when call is inactive. Flagged as nit inline — addressable in a follow-up rather than a blocker.

3 inline comments posted.

yagudaev:
Addressed both BLOCKERs in the latest commit.

1. **`file://` thumbnails broken under webSecurity** — fixed by hydrating disk-stored attachments to base64 in main on every `getAttachmentsForMessage` / `getAttachmentsForConversation` IPC. `hydrateAttachmentData` reads the file once when `storage='file'` and `data` is empty, base64-encodes it, and ships it to the renderer. The renderer's `attachmentDataUrl` is now `data:` URLs only — the file:// branch is gone, so no same-origin issues.
2. **Docs audience violation** — rewrote step 2 in `attachments.mdx` to say "Each image is saved locally with the message — you don't configure anything." Removed `local SQLite database`, `stored inline as base64`, `Electron userData/attachments`, and the 5 MB threshold.

Note: the inline/disk split is still kept in storage (smaller DB rows for >5MB images), it's just transparent to the renderer now. Memory cost on read is unchanged from before for inline; for disk-stored ones it's a one-shot read per attachment per fetch — fine for chat-scale lists.

Verified: typecheck green, 71/71 desktop tests pass, build green. (Pre-existing `archiver` failure on `diagnostic-bundle.test.ts` from origin/main is unchanged.)

PR #343, opened 2026-05-01
feat(desktop): structured tool-call panels with streaming brain response (NAN-702)

yagudaev:
Linear: [NAN-702](https://linear.app/nano3labs/issue/NAN-702)

## Why

- Today a brain tool-call shows a spinner with no detail, then dumps a wall of text into a `[Brain] …` chat bubble on completion. The wait feels stalled and the result lands disconnected from the call that produced it.
- The renderer was double-displaying the same SSE data — a global `streamingText` bubble plus the `ThinkingDots` block plus the tool row — because `tool.progress` events were piped into the assistant's transcript stream.

## What

- `desktop/src/renderer/src/components/ToolCallRow.tsx` — rebuilt as three sections: header, always-expanded Parameters JSON, and a Response panel that streams in. While in-progress, the Response panel shows the latest step caption ("Searching documents…") above the streaming text. Errors take over the Response area with a red side-accent and the params stay visible. Long completed responses gain a Collapse toggle past 800 chars.
- `desktop/src/renderer/src/lib/tool-call-store.ts` — added `step?: string` to `ToolCallEntry` and a new `applyToolCallProgress(entries, callId, { textDelta?, step? })` reducer that appends deltas into `result`. `applyToolCallCompleted` no longer overwrites streamed text when the completion result is empty or matches what we already have, which kills the flicker on done.
- `desktop/src/renderer/src/lib/tool-call-store.test.ts` — new file. Covers append, step set, no-op-on-missing-id, multi-entry preservation, and the "don't blow away streamed text on completion" behavior. 24 tests total in this file.
- `desktop/src/renderer/src/lib/use-realtime.ts` — `onToolProgress(callId, summary)` is replaced by `onToolCallProgress(callId, { textDelta?, step? })`. The handler now reads `data.textDelta`, `data.step`, and falls back to legacy `data.summary` so wire compatibility is preserved.
- `desktop/src/renderer/src/pages/ChatPage.tsx` — wires the new callback into `applyToolCallProgress`, derives `inFlightBrainCall` from `toolCalls`, and gates both `streamingText` bubble and `ThinkingDots` while a brain call is mid-flight so the row owns the surface.
- `relay-server/src/tools/brain.ts` — emits `tool.progress { textDelta }` for each SSE content delta, plus `tool.progress { step }` on `step_started` and `step_complete` (legacy `summary` field kept on `step_complete` for mobile).
- `relay-server/src/types.ts` — `ToolProgressEvent` extended with optional `step` and `textDelta`; `summary` made optional.
- `docs/src/content/docs/desktop/tool-calls.mdx` — added a "What you see during a tool call" section in user-facing prose. No event names, no IPC channels.

## Test plan

- [x] `yarn workspace voiceclaw-desktop typecheck` — green.
- [x] `yarn workspace voiceclaw-desktop test` — 117 passed (was 93; +24 new in `tool-call-store.test.ts`).
- [x] `yarn workspace voiceclaw-desktop build` — green.
- [x] `yarn workspace relay-server test` — 72 passed, 2 skipped (unchanged from main).
- [ ] **Manual repro** (run on a dev box with relay + brain wired up):
  1. `yarn dev` in `desktop/`.
  2. Start a call with `brainAgent: 'enabled'`. Ask something the brain has to research (e.g. "find the most recent issue in the inbox").
  3. Verify the tool row appears with a spinning icon, Parameters panel shows the JSON query, Response panel shows the step caption ("Searching documents…") and starts filling with streamed text.
  4. Verify there is no `ThinkingDots` block and no separate `streamingText` bubble while the row is streaming.
  5. On completion, verify the step caption disappears, the streamed text freezes (no flicker, no duplicated text), the row turns green, and the duration is stamped.
  6. Force an error path (e.g. point `BRAIN_GATEWAY_URL` at an unreachable host) and verify the Response area is replaced with the error message and the row gains a red side-accent while Parameters stay expanded.

## Screenshots

I did not capture live screenshots — running the Electron renderer through Playwright on the bundled relay+openclaw stack isn't a one-shot from this worktree, and the task explicitly allows skipping if not practical. Manual repro plan above is the verification surface.

## Caveats

- The relay is now sending one `tool.progress` event per SSE content delta. For a 2 KB brain reply that is ~20–40 events. Each is a tiny JSON message and the existing `streamingText` codepath was already handling per-token transcript deltas at higher rates, so this should be a non-event for throughput, but worth keeping an eye on once it lands on a long-running call.
- Mobile (`mobile/lib/use-realtime.ts`) still reads the legacy `data.summary` field for its `onToolProgress` callback. The relay still emits `summary` alongside the new `step` field, so mobile is unaffected by this PR.

### DIFF
--- a/desktop/src/renderer/src/components/ToolCallRow.tsx
+++ b/desktop/src/renderer/src/components/ToolCallRow.tsx
@@ -2,14 +2,15 @@ import { useEffect, useRef, useState } from 'react'
 import { ChevronDown, ChevronRight, Loader2, CheckCircle2, XCircle, Ban } from 'lucide-react'
 import type { ToolCallEntry } from '../lib/tool-call-store'
 
+const RESPONSE_COLLAPSE_THRESHOLD = 800
+
 interface ToolCallRowProps {
   entry: ToolCallEntry
 }
 
 export function ToolCallRow({ entry }: ToolCallRowProps) {
-  const { status, name, args, result, error, startedAt, durationMs } = entry
-  const [argsExpanded, setArgsExpanded] = useState(false)
-  const [resultExpanded, setResultExpanded] = useState(status === 'error')
+  const { status, name, args, result, error, startedAt, durationMs, step } = entry
+  const [responseCollapsed, setResponseCollapsed] = useState(false)
   const [elapsed, setElapsed] = useState(0)
   const intervalRef = useRef<ReturnType<typeof setInterval> | null>(null)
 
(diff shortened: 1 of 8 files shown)


desktop/src/renderer/src/components/ToolCallRow.tsx
```diff
@@ -2,14 +2,15 @@ import { useEffect, useRef, useState } from 'react'
 import { ChevronDown, ChevronRight, Loader2, CheckCircle2, XCircle, Ban } from 'lucide-react'
 import type { ToolCallEntry } from '../lib/tool-call-store'
 
+const RESPONSE_COLLAPSE_THRESHOLD = 800
+
 interface ToolCallRowProps {
   entry: ToolCallEntry
```
yagudaev, line 8: **Bug — empty `textDelta` silently skipped, but a more serious issue is missing referential stability for unaffected rows.**

`applyToolCallProgress` returns a new array on every call (via `.map()`). Unaffected entries preserve their object identity (`return e`), which is good — but `ToolCallRow` is not wrapped in `React.memo`. That means every progress event (20–40 per brain reply) re-renders every mounted `ToolCallRow`, not just the one receiving deltas. For the current scale (few rows) this is fine, but wrapping `ToolCallRow` in `memo` would cost nothing and future-proof it.

desktop/src/renderer/src/lib/tool-call-store.ts
```diff
@@ -24,15 +30,37 @@ export function applyToolCallStarted(
...
   durationMs: number,
   result: string,
 ): ToolCallEntry[] {
-  return entries.map((e) =>
-    e.callId === callId ? { ...e, status: 'success', durationMs, result } : e,
-  )
+  return entries.map((e) => {
+    if (e.callId !== callId) return e
+    const finalResult =
+      result && result.length > 0 && result !== e.result ? result : (e.result ?? result)
```
yagudaev: **Logic gap — trailing-newline mismatch causes flicker on completion.**

```ts
const finalResult =
  result && result.length > 0 && result !== e.result ? result : (e.result ?? result)
```

If the relay's `tool_call.completed` result has a trailing newline that the streamed deltas don't (or vice versa), `result !== e.result` is `true` and the completion overwrites the streamed text with a visually identical string — causing a flicker. The PR description acknowledges a "don't blow away streamed text" goal; the comparison should be trim-normalised:

```ts
const finalResult =
  result && result.trim() && result.trim() !== (e.result ?? '').trim()
    ? result
    : (e.result ?? result)
```

The existing test suite does not cover this edge case.

desktop/src/renderer/src/lib/use-realtime.ts
```diff
@@ -1,9 +1,11 @@
-// useRealtime — WebSocket hook for relay server connection
-// Port of mobile/lib/use-realtime.ts, using AudioEngine instead of native modules
-
 import { useCallback, useEffect, useRef, useState } from 'react'
 import { AudioEngine } from './audio-engine'
 
+export interface ToolCallProgressDelta {
+  textDelta?: string
+  step?: string
+}
```
yagudaev: **DRY violation — `ToolCallProgressDelta` declared twice.**

`ToolCallProgressDelta` is declared here in `use-realtime.ts` (line 4–7) and again in `tool-call-store.ts` (line 15–18). The shapes are identical so TypeScript is happy, but they'll drift silently if either is changed. One of them should import from the other:

```ts
// use-realtime.ts
import type { ToolCallProgressDelta } from './tool-call-store'
```

Not a blocker but will bite you on the next field addition.

desktop/src/renderer/src/pages/ChatPage.tsx
```diff
@@ -629,6 +628,11 @@ export function ChatPage() {
     [messages, toolCalls],
   )
 
+  const inFlightBrainCall = useMemo(
+    () => toolCalls.some((t) => t.status === 'in-progress' && t.name === 'ask_brain'),
+    [toolCalls],
+  )
```
yagudaev: **`inFlightBrainCall` stays `true` when a brain call errors — gates ThinkingDots/streamingText indefinitely.**

```ts
const inFlightBrainCall = useMemo(
  () => toolCalls.some((t) => t.status === 'in-progress' && t.name === 'ask_brain'),
  [toolCalls],
)
```

This is correct for the happy path. But when the brain call errors, `applyToolCallFailed` sets `status='error'`, so `inFlightBrainCall` drops to `false` correctly.

However: the companion `onBrainResult` fires a `brain.result` event from the relay on success, which is separate from `tool_call.completed`/`tool_call.failed`. If the relay fires `brain.result` _before_ `tool_call.failed` (or if `tool_call.failed` is dropped), the status stays `'in-progress'` forever and the gate never clears. Consider a timeout/fallback or verifying the event ordering guarantee in the relay.

desktop/src/renderer/src/pages/ChatPage.tsx
```diff
@@ -865,7 +869,7 @@ export function ChatPage() {
           )
         })}
         {/* Streaming text */}
-        {streamingText.trim() && (
+        {streamingText.trim() && !inFlightBrainCall && (
```
yagudaev: **Suppression of `streamingText` affects non-brain producers — BLOCKER.**

```tsx
{streamingText.trim() && !inFlightBrainCall && (
```

`streamingText` can also be set by `onTranscriptDelta` (assistant turns) and by the text-chat path (`streamTextChat / onToken`). If a `web_search` call is in flight _and_ the model simultaneously streams a transcript delta, the streaming bubble is suppressed. More critically: the text-chat path in `handleComposerSubmit` (line ~544 `setStreamingText(full)`) would silently vanish if any `in-progress` brain call exists in the list — even a stale one from a prior session.

The gate should be narrowed: suppress the streaming bubble only when the in-flight brain call is the active _cause_ of the stream, not as a global blanket. One safe approach is to also check that `streamingRole === 'assistant'` and the text originated from the realtime path (not text-chat).

desktop/src/renderer/src/lib/tool-call-store.test.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  it('preserves other entries', () => {
+    const entries: ToolCallEntry[] = [
+      baseEntry({ callId: 'call-1', result: 'a' }),
+      baseEntry({ callId: 'call-2', result: 'b' }),
+    ]
+    const next = applyToolCallProgress(entries, 'call-1', { textDelta: 'c' })
+    expect(next[0].result).toBe('ac')
+    expect(next[1].result).toBe('b')
+  })
```
yagudaev, line 88: **Missing test: empty-string `textDelta` is silently dropped.**

The store checks `if (delta.textDelta)` (truthy), so an empty-string delta is a no-op. That is the right behaviour, but there is no test for it. A delta of `''` should leave `result` unchanged — add a case to confirm that contract is intentional and stays that way.

yagudaev (commented):
## Review — requesting changes

Tests green (117 desktop, 72 relay). Typecheck clean. Docs are user-facing — no internal event names or IPC channels; audience separation **pass**. No ticket IDs in source. No file-level commit-message-paragraph comments added by this PR. Brand: warm neutrals, no AI-cyan — **pass**.

### BLOCKERs

**1. `streamingText` over-suppression — ChatPage.tsx:872**
The `!inFlightBrainCall` gate fires for _all_ `streamingText` producers: `onTranscriptDelta`, text-chat `onToken`, and any in-progress brain call including stale ones from a previous session that never received `tool_call.failed`. If text-chat is running and a prior brain call is sitting in `in-progress` state, the streaming bubble vanishes. The gate must be narrowed to only suppress when the brain call is the active source of the stream.

### Non-blocking bugs

**2. Trailing-newline flicker — tool-call-store.ts:61**
`applyToolCallCompleted` uses strict equality (`result !== e.result`) to decide whether to overwrite streamed text. A trailing newline difference between the streamed accumulation and the final result will be treated as a conflict and the streamed text will be clobbered — the exact flicker the PR claims to fix. Fix: trim-normalise both sides before comparing.

**3. `inFlightBrainCall` stuck forever if `tool_call.failed` is lost — ChatPage.tsx:634**
The gate only clears when `applyToolCallFailed`/`applyToolCallCompleted` run. If `brain.result` fires (and `onBrainResult` persists the answer to history) but `tool_call.failed` is dropped, status stays `'in-progress'` and the gate never clears. Needs an event-ordering guarantee in the relay or a client-side timeout/cleanup.

### Code quality

**4. Duplicate `ToolCallProgressDelta` — use-realtime.ts:4 and tool-call-store.ts:15**
Same interface declared in two files. They happen to be structurally identical today, but will drift silently. `use-realtime.ts` should import from `tool-call-store.ts`.

**5. Missing test: empty-string `textDelta` — tool-call-store.test.ts**
`if (delta.textDelta)` is a truthy guard — empty string is intentionally dropped. No test documents that contract.

**6. Performance: `ToolCallRow` not memoized**
Every `tool.progress` event creates a new outer array (from `.map()`), triggering a re-render of every mounted `ToolCallRow`. Unaffected entry objects are referentially stable, but without `React.memo` React re-renders all rows anyway. Wrapping `ToolCallRow` in `memo` would eliminate this at zero cost.

### Summary

One blocker (streamingText over-suppression) needs a fix before merge. Items 2–6 are non-blocking but should be tracked. Wire compatibility with mobile (legacy `summary` field kept) is correct. Docs are clean.

yagudaev:
Addressed in latest commit:

1. **BLOCKER (`ChatPage.tsx:872`)** — dropped the `!inFlightBrainCall` gate from both `streamingText` render and `ThinkingDots`. Brain progress events already land in the row via `applyToolCallProgress`, so `streamingText` only ever holds realtime-model output now — no need to suppress it, and removing the gate also fixes the secondary "stuck-true if `tool_call.failed` was missed" concern by deleting the predicate entirely.

2. **`tool-call-store.ts:61`** — extracted `pickFinalResult(streamed, completion)` that compares trimmed values, so a trailing newline on the canonical completion result keeps the streamed text instead of clobbering with a visually identical string. Added a test case for that exact path.

3. **Duplicated `ToolCallProgressDelta`** — single declaration in `tool-call-store.ts`; `use-realtime.ts` imports + re-exports.

4. **Empty-textDelta gap** — added a test case verifying `applyToolCallProgress` doesn't blow away existing `result` when `textDelta === ''`.

Skipping `React.memo` on `ToolCallRow` for this PR — happy to do it as a follow-up if profiling shows it matters at chat scale.

typecheck + 119/119 tests + build all green.

PR #345, opened 2026-05-01
feat(desktop): allow typed-text interrupt of streaming reply (NAN-704)

yagudaev:
## Why

Voice already supports interrupting the assistant mid-reply by speaking. Typed text didn't — the composer disabled itself while a reply was streaming, forcing the user to wait for the assistant to finish before they could redirect. This brings text parity with voice: type, press Enter, the in-flight reply stops, and the new turn takes over.

Linear: [NAN-704](https://linear.app/nano3labs/issue/NAN-704)

## What

- **`desktop/src/renderer/src/pages/ChatPage.tsx`**: drop `disabled={isThinking || !!streamingText}` from `<ChatComposer>`. In `handleComposerSubmit`, cancel any in-flight text stream up front and clear `streamingText` / `isThinking` before starting the new turn. The half-reply is not persisted to SQLite — `onDone` only writes when it actually fires.
- **`desktop/src/renderer/src/lib/text-chat.ts`**: the cancel fn returned from `streamTextChat` is now silent. It flips the `didFinish` guard and closes the websocket, without firing `onError('cancelled')`. This is what makes the existing `onToken` / `onDone` callbacks safe no-ops after cancel and prevents an "Error: cancelled" assistant message from landing in history.
- **`docs/src/content/docs/desktop/typing-messages.mdx`**: removed the "When the composer goes quiet" section that promised the composer would disable itself. Replaced with a section describing the new interrupt behavior, in the warm editorial voice the rest of the page uses.

## Test plan

- [ ] Outside a call, with the assistant offline, send a message that triggers a long reply.
- [ ] While the reply is streaming, type a different message and press Enter.
- [ ] Confirm: the partial reply disappears from the chat, the new user message appears, and the assistant streams a fresh reply for the second prompt.
- [ ] Confirm in History (or by reloading the conversation) that no half-reply or "Error: cancelled" assistant message was persisted — only the two user messages and the final assistant reply.
- [ ] Confirm the composer never greys out while a reply is streaming.
- [ ] `yarn workspace voiceclaw-desktop typecheck` — green
- [ ] `yarn workspace voiceclaw-desktop test` — green (100/100)
- [ ] `yarn workspace voiceclaw-desktop build` — green
- [ ] `yarn workspace voiceclaw-docs build` — green

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/desktop/src/renderer/src/lib/text-chat.ts
+++ b/desktop/src/renderer/src/lib/text-chat.ts
@@ -101,6 +101,8 @@ export function streamTextChat(
   }
 
   return () => {
-    if (!didFinish) finish('error', 'cancelled')
+    if (didFinish) return
+    didFinish = true
+    try { ws?.close() } catch {}
   }
 }
--- a/desktop/src/renderer/src/lib/text-chat.ts
+++ b/desktop/src/renderer/src/lib/text-chat.ts
@@ -71,12 +71,14 @@ export function streamTextChat(
         ws?.send(JSON.stringify({ type: 'text.input', text }))
         break
       case 'transcript.delta':
+        if (didFinish) break
         if (data.role === 'assistant' && typeof data.text === 'string') {
           fullText += data.text
           callbacks.onToken(fullText)
         }
         break
       case 'transcript.done':
+        if (didFinish) break
         if (data.role === 'assistant' && typeof data.text === 'string') {
           fullText = data.text
           sawDone = true
@@ -84,9 +86,11 @@ export function streamTextChat(
         }
         break
       case 'turn.ended':
+        if (didFinish) break
         if (!sawDone) finish('done', fullText)
         break
       case 'error':
+        if (didFinish) break
         finish('error', data.message || `relay error ${data.code ?? ''}`.trim())
         break
     }
--- a/desktop/src/renderer/src/pages/ChatPage.tsx
+++ b/desktop/src/renderer/src/pages/ChatPage.tsx
@@ -476,6 +476,12 @@ export function ChatPage() {
   }, [toggleMute, endCall])
 
   const handleComposerSubmit = useCallback(async (text: string) => {
+    if (textChatCancelRef.current) {
+      textChatCancelRef.current()
+      textChatCancelRef.current = null
+      setStreamingText('')
+      setIsThinking(false)
+    }
     const convId = await ensureConversation()
     const persisted = await addMessage(convId, 'user', text)
     setTypedMessageIds((prev) => {
@@ -525,7 +531,6 @@ export function ChatPage() {
     setIsThinking(true)
     streamingRoleRef.current = 'assistant'
     setStreamingRole('assistant')
-    textChatCancelRef.current?.()
     textChatCancelRef.current = streamTextChat(text, {
       serverUrl,
       apiKey,
@@ -936,7 +941,6 @@ export function ChatPage() {
       <ChatComposer
         onSubmit={handleComposerSubmit}
         onAttach={handlePickImage}
-        disabled={isThinking || !!streamingText}
       />
 
       {/* Call controls */}
--- a/docs/src/content/docs/desktop/typing-messages.mdx
+++ b/docs/src/content/docs/desktop/typing-messages.mdx
@@ -64,8 +64,8 @@ The chat window's existing shortcuts still work alongside the composer:
 
 The composer respects the OS clipboard. Cmd+V pastes the exact text from your clipboard, including hidden characters like tabs and newlines — there's no normalization, sanitization, or auto-formatting between the clipboard and the assistant.
 
-## When the composer goes quiet
+## Interrupting a streaming reply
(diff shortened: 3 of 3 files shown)


desktop/src/renderer/src/lib/text-chat.ts
```diff
@@ -101,6 +101,8 @@ export function streamTextChat(
   }
 
   return () => {
-    if (!didFinish) finish('error', 'cancelled')
```
yagudaev, line 104: **Late-token race: `onToken` has no `didFinish` guard.**

`finish()` checks `if (didFinish) return` before invoking `onDone`/`onError`. The `transcript.delta` case (file line 76) calls `callbacks.onToken(fullText)` directly, bypassing that guard.

After `cancel()` sets `didFinish = true` and calls `ws.close()`, any `transcript.delta` frames already in the WebSocket receive buffer will still dispatch through `onmessage` and call `setStreamingText` — briefly overwriting the new turn's streaming text with stale content from the cancelled stream.

Fix:
```ts
case 'transcript.delta':
  if (didFinish) break   // suppress late buffered deltas after cancel
  if (data.role === 'assistant' && typeof data.text === 'string') {
    fullText += data.text
    callbacks.onToken(fullText)
  }
  break
```

yagudaev (commented):
## Review: request-changes

**Typecheck:** green. **Tests:** 100/100 green.

---

### BLOCKER — late `onToken` after cancel (text-chat.ts line 76)

The cancel fn correctly sets `didFinish = true` before calling `ws.close()`, which prevents `onDone` and `onError` (both routed through `finish()`) from firing after cancel. But `onToken` inside the `transcript.delta` case calls `callbacks.onToken(fullText)` *directly* — it never checks `didFinish`. WebSocket implementations queue received-but-not-yet-dispatched frames; any `transcript.delta` frame buffered before the socket fully closes will still reach `onmessage`, accumulate into `fullText`, and call `setStreamingText` on the new (post-cancel) render cycle. The `fullText` accumulation is harmless (it is closure-local to the cancelled stream), but the `setStreamingText` side-effect is shared UI state and will briefly flash stale text from the dead turn over the new turn.

Fix — add a guard at the top of the `transcript.delta` case, or at the top of `onmessage`:

```ts
case 'transcript.delta':
  if (didFinish) break   // suppress late-arriving deltas after cancel
  if (data.role === 'assistant' && typeof data.text === 'string') {
    fullText += data.text
    callbacks.onToken(fullText)
  }
  break
```

---

### Gap — no tests for silent-cancel behavior (text-chat.ts)

There is no `text-chat.test.ts`. The PR removes `onError("cancelled")` and adds a silent cancel, which is behaviorally observable (no error message persisted, no flash). This path should have a vitest test: mock `WebSocket`, call the cancel fn mid-stream, assert `onDone`/`onError` are never called and `onToken` stops firing after cancel. Not a blocker on its own, but combined with the unguarded `onToken` the lack of tests means this race has no regression coverage.

---

### Code style — clean

No semicolons, helpers at bottom, public interface at top. No ticket IDs in source comments, no file-level descriptive comment blocks. CLAUDE.md style rules: pass.

---

### Docs audience — pass

The new "Interrupting a streaming reply" section is entirely user-facing: it describes what happens and why, names no internal symbols (`streamTextChat`, `textChatCancelRef`, `didFinish`, IPC channels), and matches the warm editorial voice of the rest of the page.

---

### Logic review — otherwise correct

- **Cancel-before-persist**: `textChatCancelRef.current()` is called at the top of `handleComposerSubmit`, before `await ensureConversation()` and before `await addMessage(...)`. The old cancel at line 532 (after `setIsThinking(true)`) is correctly removed. `onDone` only writes to SQLite when it actually fires, so the half-reply is never persisted. ✓
- **State reset ordering**: `setStreamingText("")` + `setIsThinking(false)` happen synchronously before the `await` chain that leads to `setIsThinking(true)` for the new turn. React batches these so no flicker. ✓
- **In-call path**: `realtimeRef.current.sendUserText` is reached after the cancel block and is unaffected. No regression. ✓
- **Send button gating**: `ChatComposer` gates `canSubmit` on `!disabled && isSubmittable(value)`. No `disabled` prop passed from `ChatPage`, so the button is gated solely on `isSubmittable(value)`. No leftover `isThinking`/`streamingText` checks. ✓
- **`ws.onclose` after cancel**: cancel sets `didFinish = true` before calling `ws.close()`, so `onclose` sees the guard and skips `finish()`. ✓

---

**Verdict: BLOCKER — fix the unguarded `onToken` before merge.** Once that is patched (one line), add a `text-chat.test.ts` covering silent-cancel, then this is good to go.

yagudaev:
Addressed in latest commit: every `onmessage` branch (`transcript.delta`, `transcript.done`, `turn.ended`, `error`) now bails on `didFinish` so late frames after cancel can't fire `onToken`/`onDone`/`onError`. typecheck + 100/100 tests green.

PR #366, opened 2026-05-01
feat(desktop): on-screen drawing during screen-share (experimental)

yagudaev:
## Why

So I can circle things on my screen while talking to Gemini and have the model actually see what I'm pointing at — instead of using my voice to describe spatial layout. Two tools, deliberately small: **Draw** (freehand) and **Clear**. Works whether I'm sharing a full display or a single application window.

Strokes are composited into the captured JPEG **in the renderer** before it goes upstream, so it works for any source type. The trace keeps the un-annotated capture + a strokes-only PNG as sibling files next to the composite, so I can see both in the log.

## Screenshots

The UI is one button strip — `Draw` `Clear` `Stop` — that only renders while a Gemini call is active *and* a screen is being shared. Driving that combined runtime state from headless Playwright would require a real WSS session and a real screen capture source, so no captures attached. Will add screenshots after manual verification.

<details>
<summary>Test plan</summary>

- [ ] `yarn workspace voiceclaw-desktop dev` boots without main-process errors
- [ ] Start a Gemini call, share a full display — overlay window appears, click-through works
- [ ] Click **Draw** — cursor becomes crosshair, scribble draws an orange stroke over the screen
- [ ] Ask Gemini "what did I just draw on the screen?" — answer matches what was drawn
- [ ] Click **Clear** — strokes vanish from the overlay; next captured frame has no strokes
- [ ] Click **Stop** — overlay hides, draw mode resets
- [ ] Repeat the test sharing a **single window** (e.g., Chrome): strokes drawn over the window must land at the matching position in the captured frame, even after moving/resizing the window
- [ ] Per-turn trace dir under the relay's session media folder contains `NNNN.jpeg` (composite), `NNNN.original.jpeg`, and `NNNN.strokes.png` for ticks where strokes were present
- [ ] Verify VoiceClaw is still reachable via Cmd+Tab and the Dock during screen share (skipTransformProcessType regression check)
</details>

<details>

**Architecture**
- New transparent always-on-top `BrowserWindow` (`desktop/src/main/draw-overlay.ts`) sized to the shared display. Click-through by default (`setIgnoreMouseEvents(true, { forward: true })`); switches to interactive when draw mode is on.
- Renderer route `?view=draw-overlay` mounts `DrawCanvas` which captures pointer events, stores polylines in display-CSS-coords, and ships them to main, which forwards to the chat renderer.
- `ScreenCapture` accepts an `AnnotationProvider` exposing a `SourceContext`:
  - `display` kind → composite at `scale = frame_w / display_w`
  - `window` kind with `windowBounds` → subtract origin, scale by `frame_w / window_w`, points outside the window's rect are clipped on draw
- For window sources, ChatPage polls `screen:getWindowBounds(windowId)` at 1 Hz so strokes track window moves/resizes.

**Wire protocol**
- `frame.append` (relay client→server) gains an optional `annotation: { original, strokesPng }` field. Composite is forwarded to Gemini as before; sibling artifacts are saved to the per-turn dir.
- `media.onVideoFrame` writes `NNNN.original.jpeg` + `NNNN.strokes.png` alongside the canonical `NNNN.jpeg` when annotation is present.

**Window bounds**
- macOS-only path uses [`get-windows`](https://github.com/sindresorhus/get-windows) (Sindre, MIT, prebuilt Swift binary, ~313 kB). Parses the CGWindowID out of the chromeMediaSourceId (`window:<id>:<displayId>`) and resolves live screen-rect via `openWindows()`.
- Polling interval: 1 s. Lookup failures fall back to display-scaling (best-effort) without crashing.

**Non-goals for v1**
- No color picker, no eraser, no thickness control. Two tools, single rust stroke color.
- No stroke persistence across share sessions.
- No multi-display strokes (one overlay window per share).
- Title-bar offset for window captures: assumed 0. macOS `kCGWindowBounds` includes the title bar; the captured stream may or may not. If strokes look ~24 px off vertically on window shares, that's where to look.

**Codex/Gemini review lessons reused**
The screen-frame indicator PR's review (memory #3191) flagged TS-undeclared `currentDisplayId`, missing `skipTransformProcessType: true`, and unvalidated IPC payloads. All three are addressed up-front in `draw-overlay.ts`.
</details>

### DIFF
--- a/desktop/src/renderer/src/call-bar/call-bar.css
+++ b/desktop/src/renderer/src/call-bar/call-bar.css
@@ -129,18 +129,19 @@ body.call-bar-preview .call-bar {
   cursor: pointer;
 }
 
-.call-bar.is-muted .call-bar__mark {
-  color: var(--brand-ink-muted);
-}
-
 /* Mute badge --------------------------------------------------------- */
 
+/* Sits centered over the waveform region — that's the "energy" the user
+   is most likely to glance at, and replacing it with a static muted mic
+   makes the off-air state unambiguous. The brand mark above it stays at
+   normal color so the bar still reads as "on a call". */
 .call-bar__mute-badge {
   position: absolute;
-  top: 4px;
-  right: 4px;
-  width: 14px;
-  height: 14px;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  width: 18px;
+  height: 18px;
   display: flex;
   align-items: center;
   justify-content: center;
@@ -148,6 +149,14 @@ body.call-bar-preview .call-bar {
   background: var(--brand-accent);
   color: var(--brand-paper);
   pointer-events: none;
+  z-index: 2;
+}
+
+/* Hide the waveform completely while muted so the badge doesn't sit on
+   top of bouncing bars (output activity still reads through the mark
+   color shift if needed). */
+.call-bar.is-muted .call-bar__waveform {
+  visibility: hidden;
 }
 
 /* Waveform pips ------------------------------------------------------- */
--- a/desktop/src/renderer/src/call-bar/call-bar.css
+++ b/desktop/src/renderer/src/call-bar/call-bar.css
@@ -129,6 +129,27 @@ body.call-bar-preview .call-bar {
   cursor: pointer;
 }
 
+.call-bar.is-muted .call-bar__mark {
+  color: var(--brand-ink-muted);
+}
+
+/* Mute badge --------------------------------------------------------- */
+
+.call-bar__mute-badge {
+  position: absolute;
+  top: 4px;
+  right: 4px;
+  width: 14px;
+  height: 14px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  border-radius: 50%;
+  background: var(--brand-accent);
+  color: var(--brand-paper);
+  pointer-events: none;
+}
+
 /* Waveform pips ------------------------------------------------------- */
 
 .call-bar__waveform {
--- a/desktop/src/renderer/src/draw-overlay/draw-overlay.css
+++ b/desktop/src/renderer/src/draw-overlay/draw-overlay.css
@@ -1,0 +1,71 @@
+html,
+body,
+#root {
+  margin: 0;
+  padding: 0;
+  width: 100%;
+  height: 100%;
+  background: transparent;
+  overflow: hidden;
+  cursor: default;
+}
+
+.draw-overlay__canvas {
+  position: fixed;
+  inset: 0;
+  width: 100vw;
+  height: 100vh;
+  display: block;
+  background: transparent;
+  pointer-events: auto;
+}
+
+.draw-overlay__canvas--drawing {
+  cursor: crosshair;
+}
+
+.draw-overlay__toolbar {
+  position: fixed;
+  top: 24px;
+  left: 50%;
+  transform: translateX(-50%);
+  display: flex;
+  gap: 8px;
+  padding: 8px 12px;
+  border-radius: 12px;
+  background: rgba(20, 16, 12, 0.85);
+  backdrop-filter: blur(12px);
+  -webkit-backdrop-filter: blur(12px);
+  box-shadow: 0 8px 24px rgba(0, 0, 0, 0.32);
+  font-family: -apple-system, BlinkMacSystemFont, "SF Pro Text", system-ui, sans-serif;
+  font-size: 13px;
+  pointer-events: auto;
+  z-index: 10;
+}
+
+.draw-overlay__button {
+  appearance: none;
+  border: none;
+  background: transparent;
+  color: rgba(255, 255, 255, 0.85);
+  padding: 6px 12px;
+  border-radius: 8px;
+  cursor: pointer;
+  font: inherit;
+  transition: background-color 120ms ease;
+}
+
+.draw-overlay__button:hover {
+  background: rgba(255, 255, 255, 0.08);
+  color: #fff;
+}
+
+.draw-overlay__button--primary {
+  background: #FF6B3D;
+  color: #fff;
+}
+
+.draw-overlay__button--primary:hover {
+  background: #ff5a26;
+  color: #fff;
+}
--- a/docs/src/content/docs/desktop/screen-annotations.mdx
+++ b/docs/src/content/docs/desktop/screen-annotations.mdx
@@ -55,7 +55,7 @@ All four are configurable from [Settings → Keyboard Shortcuts](/desktop/keyboa
 
 Each captured frame goes through this path:
 
-1. The video element samples a frame at 1 Hz.
+1. The video element samples one frame a second.
 2. VoiceClaw scales it down to 1536 px on the long edge and paints any current strokes on top.
 3. The composited JPEG is sent to Gemini as the user video input.
 
--- a/docs/src/content/docs/desktop/screen-annotations.mdx
+++ b/docs/src/content/docs/desktop/screen-annotations.mdx
@@ -1,0 +1,77 @@
+---
+title: Drawing on your screen
+description: Circle, underline, or scribble on whatever you're sharing — the strokes are baked into the frames sent to Gemini, so the model sees exactly what you're pointing at.
+---
+
+import { Aside } from '@astrojs/starlight/components'
+
+When you're on a Gemini Live call and screen-sharing, VoiceClaw lets you draw on the shared region with two tools: **Draw** (freehand) and **Clear**. The strokes are composited directly into the captured JPEG frames before they're sent upstream, so the model sees them — there's no separate annotation channel for it to ignore.
+
+This is meant for pointing things out by gesture instead of by description. "Look at this" beats "the third item from the left in the second column."
+
+## Starting a session
+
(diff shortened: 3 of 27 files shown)


docs/src/content/docs/desktop/screen-annotations.mdx
```diff
@@ -0,0 +1,77 @@
...
+- `Ctrl + S` — toggle screen share.
+- `Ctrl + M` — toggle mute (handy mid-drawing).
+
+All four are configurable from [Settings → Keyboard Shortcuts](/desktop/keyboard-shortcuts/).
+
+## What the model sees
+
+Each captured frame goes through this path:
+
+1. The video element samples a frame at 1 Hz.
```
yagudaev: nit: just say 1 frame a second easier to understand

yagudaev:
Dual Codex + Gemini review run via `pr-flow/scripts/review.sh` against the diff at `origin/main...origin/michael/draw-on-screen` (~3.7k lines). Reviewers were also given the attention areas from the PR description (IPC sender validation, `transformStrokes` window-source math, `handleServerMessage` usage/serverContent ordering, `MODEL_CONTEXT_WINDOWS` correctness, overlay focus stealing).

## Codex

### Critical
- Stopping a screen share only hides the overlay; it does not clear the overlay renderer's `strokesRef`, so the next share can repaint and resend stale annotations as soon as the user draws again. Clear the overlay before/inside hide. (desktop/src/renderer/src/pages/ChatPage.tsx:687)
- `startScreenShare` always calls `drawOverlay.show()` without the selected display id, so `draw-overlay` falls back to the primary display. Sharing a secondary monitor/window puts the overlay and display bounds on the wrong screen, which breaks annotation alignment. (desktop/src/renderer/src/pages/ChatPage.tsx:660)

### Should fix
- Window-source stroke transforms subtract absolute window bounds from overlay-local points. On any display with a non-zero origin, strokes are shifted by the display offset; add `context.displayBounds.x/y` before subtracting `win.x/y`. (desktop/src/renderer/src/lib/screen-capture.ts:238)
- `shortcuts:set` persists the new accelerator and unregisters the old one before confirming `globalShortcut.register` succeeded, so a failed/owned binding loses the previous shortcut and leaves the DB showing a binding that is not registered. (desktop/src/main/shortcuts.ts:51)
- The shortcuts IPC handlers accept calls from any sender, but they mutate global OS shortcuts. Restrict these to the trusted main/settings renderer, similar to the draw-overlay sender checks. (desktop/src/main/shortcuts.ts:46)
- `screen:getWindowBounds` has no sender validation even though it exposes OS window geometry to any renderer with the preload API. Restrict it to the main renderer while a window share is active. (desktop/src/main/screen-capture.ts:67)
- Split the Gemini usage/logging/context-window debug changes out of the drawing PR; they alter realtime message handling and logging independently of the annotation feature, increasing the rollback/review surface. (relay-server/src/adapters/gemini.ts:616)

### Nits
_none_

## Gemini

### Critical
- Broken window-sharing stroke math on multi-monitor setups; `transformStrokes` ignores the display origin offset when mapping overlay-relative points to absolute window bounds. If the shared window is on a secondary display, strokes will be offset or clipped entirely. Should be `(p.x + context.displayBounds.x - win.x) * scaleX`. (desktop/src/renderer/src/lib/screen-capture.ts:192)

### Should fix
- Missing IPC sender validation for `call-bar:muted` and `call-bar:audio-levels` listeners in `index.ts`; any renderer (e.g., the overlay or an exploit in a webview) could broadcast these events to manipulate the call-bar UI state. (desktop/src/main/index.ts:225)
- Missing IPC sender validation for `screen:getWindowBounds`; while only returning geometry, this sensitive system information should be restricted to the main window sender. (desktop/src/main/screen-capture.ts:68)
- The `isPlausibleAccelerator` validation is too loose; it allows invalid Electron accelerators like `"Command+"` or `"Shift+"` to reach the registration logic. (desktop/src/main/shortcuts.ts:168)

### Nits
- nit: `logUsage` and `handleServerMessage` use `any` for the message payload; defining types for the Gemini Bidi protocol responses would improve maintenance. (relay-server/src/adapters/gemini.ts:614)
- nit: use of `Number.parseInt` with an explicit base 10 is redundant here given the regex `\d+` ensures a valid decimal string. (desktop/src/renderer/src/pages/ChatPage.tsx:1410)
- nit: `findModalityTokens` and `sumModalityTokens` use `any` for the `details` array; these should be typed to match the `usageMetadata` schema. (relay-server/src/adapters/gemini.ts:1225)

## Triage

- **Critical: 2 unique findings** (one shared between reviewers).
  - **Multi-display alignment** — both reviewers flag the same bug from different angles. Codex points at `startScreenShare` in `ChatPage.tsx:660` not passing the selected display id to `drawOverlay.show()`; Gemini points at `transformStrokes` in `screen-capture.ts:192` ignoring `displayBounds.x/y`. The two are likely the same root cause: window-source on a secondary display is broken.
  - **Stale strokes on re-share** (Codex only) — `strokesRef` not cleared on stop, so resuming a share repaints prior annotations.
- **Should fix: 6** — 4 IPC sender-validation gaps (call-bar listeners, `screen:getWindowBounds`, both `shortcuts:*` handlers), one accelerator-validation atomicity bug (`shortcuts:set` unregisters before confirming the new binding), one looseness in `isPlausibleAccelerator`, plus Codex's recommendation to split Gemini logging/usage changes into a separate PR.
- **Nits: 3** — all type-safety nits in the Gemini adapter and one redundant `Number.parseInt` radix.

Neither reviewer flagged the `handleServerMessage` usage/serverContent co-arrival fix or the `MODEL_CONTEXT_WINDOWS` map as problematic. Neither flagged overlay `setFocusable(true)` focus-stealing on macOS — but neither explicitly cleared it either, so treat that one as un-reviewed rather than verified-clean.

desktop/src/renderer/src/draw-overlay/draw-overlay.css
```diff
@@ -0,0 +1,71 @@
...
+  display: block;
+  background: transparent;
+  pointer-events: auto;
+}
+
+.draw-overlay__canvas--drawing {
+  cursor: crosshair;
+}
+
+.draw-overlay__toolbar {
```
yagudaev, line 27: why are we using css here instead of tailwind?

desktop/src/renderer/src/call-bar/call-bar.css
```diff
@@ -129,6 +129,36 @@ body.call-bar-preview .call-bar {
   cursor: pointer;
 }
 
+/* Mute badge --------------------------------------------------------- */
+
+/* Sits centered over the waveform region — that's the "energy" the user
+   is most likely to glance at, and replacing it with a static muted mic
+   makes the off-air state unambiguous. The brand mark above it stays at
+   normal color so the bar still reads as "on a call". */
+.call-bar__mute-badge {
```
yagudaev, line 138: again with all this css, should be tailwind

desktop/src/renderer/src/draw-overlay/draw-overlay.css
yagudaev: These two views (call-bar + draw-overlay) deliberately don't import Tailwind — see `desktop/src/renderer/src/main.tsx:9-12`:

> Both stylesheets and the app/call-bar modules are dynamic-imported so the call-bar window doesn't drag the Tailwind build (34 kB of CSS plus the main app's tree)…

So switching to Tailwind here means accepting that ~34 kB cost on every overlay/call-bar window load. With Tailwind 4's purging it's probably much less in practice — but it's a real architectural change, not a drop-in swap.

Three options:

1. **Keep CSS** for these two views (matches the documented reason — small bundle for tiny always-on-top windows).
2. **Convert these two views to Tailwind** and drop the architectural note in main.tsx. Means importing tailwind into both view entries.
3. **Inline styles** on the two new elements I added (mute badge, draw-overlay toolbar) without converting the existing CSS files.

I'd lean (1) since it's following the existing pattern — but if you want the consistency more than the bundle savings, (2) is the clean answer. Let me know which.

desktop/src/renderer/src/call-bar/call-bar.css
yagudaev: These two views (call-bar + draw-overlay) deliberately don't import Tailwind — see `desktop/src/renderer/src/main.tsx:9-12`:

> Both stylesheets and the app/call-bar modules are dynamic-imported so the call-bar window doesn't drag the Tailwind build (34 kB of CSS plus the main app's tree)…

So switching to Tailwind here means accepting that ~34 kB cost on every overlay/call-bar window load. With Tailwind 4's purging it's probably much less in practice — but it's a real architectural change, not a drop-in swap.

Three options:

1. **Keep CSS** for these two views (matches the documented reason — small bundle for tiny always-on-top windows).
2. **Convert these two views to Tailwind** and drop the architectural note in main.tsx. Means importing tailwind into both view entries.
3. **Inline styles** on the two new elements I added (mute badge, draw-overlay toolbar) without converting the existing CSS files.

I'd lean (1) since it's following the existing pattern — but if you want the consistency more than the bundle savings, (2) is the clean answer. Let me know which.

docs/src/content/docs/desktop/screen-annotations.mdx
yagudaev: Done — changed to 'one frame a second' on https://github.com/yagudaev/voiceclaw/blob/michael/draw-on-screen/docs/src/content/docs/desktop/screen-annotations.mdx#L58. Will commit with the rest of the review-fix batch.

PR #389, opened 2026-05-03
fix(desktop): pre-schedule TTS chunks on audio thread to eliminate boundary clicks (NAN-706)

yagudaev:
## Why

Gemini TTS playback was choppy with audible crackle / static at chunk boundaries. Grok was clean. Spike in #346 pinned this to a renderer underrun: each delta was wrapped in its own AudioBufferSourceNode and chained via onended -> drainPlaybackQueue. onended fires on the main thread, so the next buffer can only .start() after a thread hop, and the gap fills with silence at 24 kHz -> a click on every chunk boundary. Gemini emits ~20-50 chunks/s vs Grok's larger chunks, hence Gemini-only.

## What

`desktop/src/renderer/src/lib/audio-engine.ts`
- Track `nextStartTime` on the engine. Each `playAudio` call schedules `source.start(Math.max(currentTime, nextStartTime))` and advances `nextStartTime += buffer.duration`. The audio thread now stitches contiguous buffers sample-accurately — no more main-thread hop between chunks.
- Live sources tracked in `Set<AudioBufferSourceNode>`. `source.onended` only removes from the set; it does not drive the next start.
- `stopPlayback` resets `nextStartTime = 0` and stops every live source, so barge-in cuts immediately and the next turn starts at `currentTime` instead of a stale schedule.
- Edge case: if the queue drains and a new chunk arrives later, `nextStartTime` is in the past, and `Math.max` clamps to `currentTime`. Covered by test.
- Removed the file-top descriptive comment (per CLAUDE.md).

`desktop/src/renderer/src/lib/audio-engine.test.ts` (new, 6 tests)
- First chunk schedules at `currentTime`.
- Three contiguous chunks schedule at monotonically increasing start times matching cumulative duration.
- After a gap, the next chunk clamps to `currentTime` rather than a stale `nextStartTime`.
- `stopPlayback` stops all live sources and resets the schedule.
- `stopPlayback` tolerates already-stopped sources without throwing.
- Natural `onended` removes a source from the live set so `stopPlayback` only stops still-playing sources.

## Test plan

- [x] `yarn workspace voiceclaw-desktop typecheck` — green
- [x] `yarn workspace voiceclaw-desktop test` — 125 passed (was 119, 6 new in audio-engine.test.ts)
- [x] `yarn workspace voiceclaw-desktop build` — green
- [ ] Manual: start a Gemini call, listen for crackle/static at chunk boundaries — should be gone
- [ ] Manual: switch to Grok, no regression
- [ ] Manual: barge-in mid-reply, start a new turn — should not have a stale audio gap

## Reference

Spike findings: #346

### DIFF
--- a/desktop/src/renderer/src/lib/audio-engine.test.ts
+++ b/desktop/src/renderer/src/lib/audio-engine.test.ts
@@ -1,0 +1,156 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { AudioEngine, SAMPLE_RATE } from './audio-engine'
+
+type FakeSource = {
+  buffer: { duration: number } | null
+  connect: ReturnType<typeof vi.fn>
+  start: ReturnType<typeof vi.fn>
+  stop: ReturnType<typeof vi.fn>
+  onended: (() => void) | null
+}
+
+type FakeGain = { gain: { value: number }, connect: ReturnType<typeof vi.fn> }
+
+type FakeContext = {
+  currentTime: number
+  destination: object
+  createBuffer: (channels: number, length: number, rate: number) => { duration: number, copyToChannel: ReturnType<typeof vi.fn> }
+  createBufferSource: () => FakeSource
+}
+
+let ctx: FakeContext
+let gain: FakeGain
+let createdSources: FakeSource[]
+
+beforeEach(() => {
+  createdSources = []
+  ctx = makeFakeContext()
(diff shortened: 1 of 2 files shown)


yagudaev:
Addressed in d82d009: dropped the two-line comment that explained old onended behavior / contrast — pure code now, no bug-history framing. typecheck + 125/125 tests still green.

PR #391, opened 2026-05-04
feat(desktop): voice preview button in Settings (NAN-713)

yagudaev:
## Why

The Settings voice picker showed names only, with no way to hear a voice before committing — the same affordance already exists in onboarding (`StepIdentity`), so users had to either remember from setup or change-and-listen-then-revert. Closes [NAN-713](https://linear.app/nano3labs/issue/NAN-713/settings-voice-picker-has-no-preview-button).

## Approach

- **Preview is a separate action.** A small play button sits beside the voice label; clicking it plays the sample without touching the selection. Errors render inline beneath the grid.
- **Samples are static, generated once per voice.** Per the ticket's hard requirement: a new `getCachedVoicePreview` helper writes the first response to `userData/voice-previews/v1/<voice>.json` and serves cached bytes on every subsequent click — no network call, works offline once warmed.
- **Onboarding still uses dynamic, name-personalized preview text** ("Hi, I'm Pam.") via the existing `speakPreview` IPC. Settings uses the cached static-text path. Audio decoding is shared via a new `lib/voice-preview.ts` module (extracted from the duplicated PCM→WAV helpers in `StepIdentity.tsx`).
- XAI voices show no preview button (the existing TTS endpoint is Gemini-only).

- [x] `yarn test` — 13 files, 124 tests pass (5 new tests cover cache hit/miss, missing-key, and per-voice isolation)
- [ ] Manual: open Settings on Gemini model → click ▶ on each voice → hear sample, selection unchanged
- [ ] Manual: click ▶ twice quickly → no overlap, second click restarts cleanly
- [ ] Manual: clear Gemini key → cached voices still preview; uncached voice surfaces "No Gemini key configured." inline
- [ ] Manual: switch model to Grok → no preview buttons render, no errors

</details>

- `getCachedVoicePreview` is in `desktop/src/main/identity.ts` next to the existing `speakGreetingPreview`. Cache version is encoded in the directory name (`v1`) so future static-text changes can bust it without manual cleanup.
- The shared decode helper is in `desktop/src/renderer/src/lib/voice-preview.ts`. `StepIdentity.tsx` was refactored to use it, deleting ~50 lines of duplicated PCM→WAV code.
- I went with **option 1 from the ticket** (port the handler) plus a tiny shared decode helper. Extracting a full `<VoicePicker>` would have meant either compromising the onboarding row layout (preview + select on one line, monospace label) or the Settings 2-col grid — not worth it for two consumers.

</details>

### DIFF
--- a/desktop/src/main/identity.test.ts
+++ b/desktop/src/main/identity.test.ts
@@ -93,3 +93,108 @@ describe('readAgentIdentity', () => {
     expect(id.voice).toBe('Charon')
   })
 })
+
+describe('getCachedVoicePreview', () => {
+  let fetchCalls: { url: string; init?: RequestInit }[] = []
+
+  beforeEach(() => {
+    writes.length = 0
+    fileSystem.clear()
+    fetchCalls = []
+    vi.doMock('electron', () => ({
+      app: {
+        getPath: () => '/tmp/voiceclaw-identity-test',
+      },
+      net: {
+        fetch: async (url: string, init?: RequestInit) => {
+          fetchCalls.push({ url, init })
+          return {
+            ok: true,
+            json: async () => ({
+              candidates: [
+                {
+                  content: {
+                    parts: [
+                      {
+                        inlineData: {
+                          data: 'BASE64DATA',
+                          mimeType: 'audio/L16;rate=24000',
+                        },
+                      },
+                    ],
+                  },
+                },
+              ],
+            }),
+            text: async () => '',
+          }
+        },
+      },
+    }))
+  })
+
+  afterEach(() => {
+    vi.resetModules()
+    vi.doUnmock('electron')
+  })
+
+  it('generates and caches the preview on first request', async () => {
+    const { getCachedVoicePreview } = await import('./identity')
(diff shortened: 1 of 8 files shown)


yagudaev:
## Review responses

Both Codex and Gemini reviewed `5ad1648`. All Critical + Should-fix items addressed in `617f543`:

| Finding | Reviewer | Status |
|---|---|---|
| Blob URLs not revoked → memory leak | Codex (should-fix) + Gemini (critical) | Fixed — `decodeVoicePreviewAudio` now returns `{ audio, revoke }`; both callers track + invoke `revoke` on end/error/replace/unmount |
| Sync FS in main-process IPC handler | Gemini (should-fix) | Fixed — switched to `node:fs/promises` for the cache helpers |
| Race condition in `handleVoicePreview` | Gemini (should-fix) | Fixed — monotonic `previewTokenRef` discards stale IPC results in both `SettingsPage` and `StepIdentity` |
| No request serialization for same voice | Gemini (should-fix) | Fixed — per-voice `Map<string, Promise<…>>` coalesces concurrent first-clicks; new test covers it |

Nits skipped (comment trim, i18n placeholder).

PR #393, opened 2026-05-04
feat(desktop): xAI voice previews via bundled WAVs (NAN-714)

yagudaev:
## Why

NAN-713 shipped a preview button for Gemini voices in Settings. xAI voices were left without — that ticket relied on Gemini's HTTP TTS endpoint, which xAI doesn't expose. Closes [NAN-714](https://linear.app/nano3labs/issue/NAN-714/settings-voice-picker-has-no-preview-button-for-xai-voices).

## Approach

- **Pre-record once via Realtime WS.** New `desktop/scripts/generate-xai-voice-previews.mjs` opens an xAI Realtime session per voice, asks for "Hi, I'm here.", concatenates the PCM frames, wraps them in a WAV header, writes `desktop/resources/voice-previews/xai/<voice>.wav`. Re-run only when the voice list or model changes.
- **Bundle the WAVs.** `electron-builder.yml` already maps `desktop/resources/**` → `<App>/Contents/Resources/`, so the new directory ships with no config change. ~150KB × 5 ≈ 750KB total.
- **Runtime dispatch by provider.** `getCachedVoicePreview` now checks `providerForVoice(voice)` — xAI reads from the bundled WAV (`audio/wav`, base64), Gemini keeps its existing on-demand TTS + userData cache. xAI path needs no network and no API key.
- **UI: drop the `isGemini` gate** in `SettingsPage`. Play button renders for every voice now; the inline error banner from NAN-713 covers any "preview missing" case.

- [x] `yarn test` — 13 files, 127 tests pass (2 new: bundled-WAV happy path + missing-asset error)
- [x] `node scripts/generate-xai-voice-previews.mjs` — produced 5 valid `RIFF (little-endian) data, WAVE audio, Microsoft PCM, 16 bit, mono 24000 Hz` files (verified with `file`); `afplay` exits 0 on each
- [ ] Manual: open Settings on Grok → click ▶ on each xAI voice → hear the same "Hi, I'm here." sample
- [ ] Manual: switch to Gemini → previews still work (no regression on the existing path)
- [ ] Manual: clear all API keys → xAI previews still play (bundled, no key needed)

</details>

- The script intentionally sends `turn_detection: null` and an empty `tools` array — we never feed audio in, and we don't want the model wandering into a tool call. Instructions tell it to read the user message verbatim once and stop.
- I considered shipping no samples and lazily generating into `userData` like the Gemini path. Rejected: the xAI Realtime WS is much heavier than a single HTTP POST (model load, session config round-trip, frame stream), and the ticket's hard requirement is "static files generated once" — bundling is closer to the spirit.
- xAI WAV path is read every click rather than memo-cached. Files are ~150KB each, fully on local disk; the disk read disappears in the IPC round-trip noise.

</details>


yagudaev:
## Review responses

Both Codex and Gemini reviewed `e09fc2f`. Critical: 0. Should-fix addressed in `a7df06f`:

| Finding | Reviewer | Status |
|---|---|---|
| `ws` not declared in `desktop/package.json` (works only via hoisted transitive) | Codex + Gemini | Fixed — added `ws: ^8.18.2` to desktop devDependencies; lockfile updated |
| Empty audio response hangs script for 30s timeout | Gemini | Fixed — `response.done` with `frames.length === 0` now errors immediately |
| `app.getAppPath()` may differ in dev | Gemini | No change — uses the same pattern as `tray.ts:182` and other resource lookups; verified by running the script + reading the WAVs locally |
| Verify xAI Realtime schema matches real API | Gemini | Verified by running — produced 5 valid 24kHz/16-bit WAVs (`file`-checked, `afplay` clean) |

Nits skipped (hardcoded voice list, redundant socket close, mixed fs imports — all justified inline in the commit message).

PR #395, opened 2026-05-04
fix(desktop): bundle Gemini voice previews; rip out lazy cache (NAN-715)

yagudaev:
## Why

The Gemini voice previews shipped in NAN-713 used a lazy userData cache — first click hit `gemini-3.1-flash-tts-preview:generateContent`, then persisted to disk. In practice the free-tier preview model rate-limits aggressively; users saw `TTS 429: You exceeded your current quota` instead of audio. The original ticket spec said "static files, generated once" — I shipped the wrong shape. NAN-714 already did this correctly for xAI; this PR realigns Gemini.

Closes [NAN-715](https://linear.app/nano3labs/issue/NAN-715/bundle-gemini-voice-previews-as-static-wavs-drop-lazy-userdata-cache).

## Approach

- **Pre-generate once, commit the WAVs.** New `desktop/scripts/generate-gemini-voice-previews.mjs` calls Gemini TTS for each of the 8 voices, wraps PCM16 in a WAV header, writes `desktop/resources/voice-previews/gemini/<Voice>.wav`. Idempotent (skips voices whose file already exists) and retries on 429 honoring the API's `RetryInfo.retryDelay` — necessary because we hit the limit partway through.
- **Bundle**: `electron-builder.yml` already maps `desktop/resources/**` → `<App>/Contents/Resources/`. No config change.
- **Simplify the runtime.** `getCachedVoicePreview` now just dispatches by `providerForVoice` and reads `resources/voice-previews/<provider>/<voice>.wav`. Deleted: `inFlightVoicePreviews` map, `generateAndCacheVoicePreview`, `readVoicePreviewCache`, `writeVoicePreviewCache`, `CACHE_VERSION`, `STATIC_PREVIEW_TEXT`, the `apiKey` parameter. The IPC handler no longer threads `getProviderKey('gemini')`.
- Onboarding's name-personalized "Hi, I'm Pam." preview (`speakGreetingPreview`) is untouched — that one *needs* the dynamic name and stays on the live API.

- [x] `yarn test` — 13 files, 124 tests pass; new tests assert `net.fetch` is never invoked from the preview path
- [x] Generation script: produced 8 valid 24kHz/16-bit/mono WAVs (`file`-checked); script auto-resumed past the 429 we hit during generation
- [ ] Manual: Settings → click ▶ on each Gemini voice → hear the sample (no 429, no key requirement)
- [ ] Manual: clear all API keys → previews still play
- [ ] Manual: switch to Grok → xAI previews still play (regression check on NAN-714)

</details>

- Bundle size: 8 Gemini × ~85KB ≈ 680KB. Plus xAI's 750KB from NAN-714. ~1.4MB total for voice previews — fine.
- Stale userData cache from prior installs: not cleaned up. Harmless (it just sits unread under `userData/voice-previews/v1/`). Adding a one-time cleanup would be more code than it's worth; the directory will get garbage-collected by the OS or the user can clear it.

</details>


yagudaev:
## Review responses

Both reviewers reviewed `08…` (initial commit). Critical: 1 (both flagged the same item). Should-fix: 3 unique. Addressed in the follow-up commit:

| Finding | Reviewer | Severity | Status |
|---|---|---|---|
| Source comment references ticket ID + narrates the previous quota bug | Codex + Gemini | Critical | Fixed — header rewritten as plain "what + how to run", ticket context lives in the PR body |
| Stale WAVs silently survive after a model/prompt change | Codex + Gemini | Should fix | Fixed — added `--force` flag; default still skips existing files so partial runs resume |
| `getCachedVoicePreview` is misleadingly named (nothing is cached anymore) | Gemini | Should fix | Fixed — renamed to `getBundledVoicePreview` everywhere (function, IPC import, tests) |
| `parseRetryDelayMs` greps JSON via regex | Gemini | Should fix | Fixed — uses `JSON.parse` and walks `error.details[].retryDelay` |

Nits skipped (voice-list link comment, pcmToWav magic numbers, breaking-change call-out — only one caller exists and it's already updated).

PR #397, opened 2026-05-04
fix(relay): drop stale audio queue on Gemini 1007 close to break reconnect loop

yagudaev:
## Why

Gemini Live occasionally closes the WS with `1007 Request contains an invalid argument` at user-input boundaries. The relay reconnects with the resumption handle and replays its pending audio queue — but those chunks are exactly what tripped 1007. Replaying them re-trips 1007, putting the call into a tight reconnect loop (real-world repro: close 1007 → reconnect → flush 4 audio → close 1007 → ...).

This change gates the audio drop to **close code 1007 only**. For 1006 / 1011 / 1012 / 1013 the queued audio is harmless and dropping it would lose user words, so the existing replay behavior is preserved. Control messages (`toolResponse`, `clientContent`) still flush on 1007 so the model state stays coherent.

## Test plan

- [x] `yarn workspace relay-server test test/adapters/` — all 88 tests pass (86 baseline + 2 new)
- [x] `yarn workspace relay-server test` — full suite green (126 pass, 3 skipped, no new skips)
- [x] `yarn workspace relay-server build` — `tsc` clean
- [ ] Manual: trigger a 1007 in a real call (or have it happen organically) and confirm the resumed session does not immediately close again with 1007

## Implementation notes

<details>

- New private field `dropPendingAudioOnFlush: boolean` on `GeminiAdapter`. One-shot flag, set in `handleUpstreamClose(code)` when `code === 1007`, consumed and cleared in `flushPending`.
- Set point is **after** the existing early-return guards (`!RECONNECTABLE_CLOSE_CODES.has(code) || !this.resumptionHandle`), so the flag is only set when we actually decide to reconnect.
- Consume point is `flushPending` — runs after `setupComplete` on the resumed socket. Control queue still flushes; audio queue is logged-and-dropped.
- Did not touch:
  - Video drop path (already drops every reconnect for the same 1007 reason).
  - The `goAway` → close path (`isReconnecting === true` makes `handleUpstreamClose` return early before our check, which is correct — goAway-driven rotations are graceful and the queued audio is safe).
  - The post-resume "poisoned handle" watchdog (`armPostResumeWatchdog`) — that's a different recovery path (force-fresh on no-generation), still useful as a fallback when the 1007 turns out not to be audio-shaped.
  - `openai.ts` — Gemini-specific.
- Two new tests in `gemini-reconnect.test.ts` under `send queue during rotation`:
  - 1007 close → audio queue drops to 0 on flush, control still flushes, no client-facing error.
  - 1011 close → audio queue still flushes (regression guard for the gating).

### Things noticed in passing

- `RECONNECTABLE_CLOSE_CODES` already includes `1007` with a comment hypothesizing stale **video** frames as the cause. The bug log here suggests stale **audio** can also do it (matches the `reference_tracer_locations.md` note). Worth tightening the comment in a follow-up.
- `flushPending` flushes **all** queued audio after setup. With `MAX_PENDING_AUDIO = 50` (~1s at 50Hz), that's potentially a 1s burst into a freshly resumed session. Not part of this fix, but if 1007s persist after this change, the next thing to look at is whether the burst itself is the trigger (rate, or boundary alignment with the resumption checkpoint).

</details>

### DIFF
--- a/.github/workflows/brain-e2e.yml
+++ b/.github/workflows/brain-e2e.yml
@@ -19,6 +19,9 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Enable corepack
+        run: corepack enable
+
       - uses: actions/setup-node@v6
         with:
           node-version: '22'
--- a/relay-server/src/adapters/gemini.ts
+++ b/relay-server/src/adapters/gemini.ts
@@ -82,6 +82,12 @@ export class GeminiAdapter implements ProviderAdapter {
   private pendingAudio: string[] = []
   private pendingVideo: string[] = []
   private pendingControl: string[] = []
+  // Set when the previous upstream closed with 1007 (invalid argument). The
+  // queued audio captured before that close is the most likely culprit, so
+  // replaying it into the resumed session re-trips 1007 and locks us into a
+  // reconnect loop. Control messages still flush — the model needs them to
+  // stay coherent.
+  private dropPendingAudioOnFlush = false
 
   // Per-turn latency marks, emitted on turnComplete. Gemini Live has no
   // explicit "end-of-speech" event — we proxy it via the last inputTranscription
@@ -287,6 +293,8 @@ export class GeminiAdapter implements ProviderAdapter {
       })
       return
     }
+
+    if (code === 1007) this.dropPendingAudioOnFlush = true
 
     void this.reconnect(`close code ${code}`)
   }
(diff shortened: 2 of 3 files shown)


yagudaev:
Dual review (Codex + Gemini) on the diff. Both ran against the +62/-3 patch.

## Codex

### Critical
- \`dropPendingAudioOnFlush\` drops the whole audio queue on the next flush, but for a close-triggered reconnect that queue is populated only after \`isReconnecting\` is set, so this silently discards fresh user speech captured during recovery rather than audio queued before the 1007. Clear only already-stale queued audio, or track a cutoff/generation so post-close audio can still flush. (relay-server/src/adapters/gemini.ts:294-296, 839-851)

### Should fix
_none_

### Nits
_none_

## Gemini

### Critical
_none_

### Should fix
- Use a named constant for WebSocket close codes \`1007\` (Invalid Frame Payload Data / Invalid Argument) and \`1011\` (Server Internal Error) instead of magic numbers. This improves readability and makes it easier to track where specific protocol behaviors are handled. (\`relay-server/src/adapters/gemini.ts:295\`, \`relay-server/test/adapters/gemini-reconnect.test.ts:330\`, \`349\`)

### Nits
- nit: The log message \`Flushing reconnect queue\` reports \`0 audio\` when \`dropAudio\` is true, which might be slightly confusing when immediately followed by a log saying audio chunks were dropped. Consider refining the log to reflect that audio is being intentionally suppressed (e.g., \`... (X control, 0 audio [suppressed due to 1007])\`). (\`relay-server/src/adapters/gemini.ts:849\`)
- nit: Group the reset \`this.dropPendingAudioOnFlush = false\` with the other property clears (\`this.pendingControl = []\`, etc.) at the top of the \`flush\` method to keep state management centralized and consistent. (\`relay-server/src/adapters/gemini.ts:843\`)

## Triage

**Codex's Critical is worth a closer look but I believe it's a misread, not a blocker.** Reasoning:

- \`pendingAudio\` is only ever populated while \`isReconnecting\` is true (see \`sendUpstream\` at gemini.ts:801-806). Outside that window, audio is sent directly to the upstream socket.
- So Codex is correct that the audio dropped at flush time was captured *during* the reconnect window, not strictly "before the 1007." The comment on the field (\"queued audio captured before that close\") is slightly misleading on that point.
- However, the behavior is still right: on a 1007 the mic keeps streaming continuously into the reconnect queue, and whatever payload tripped 1007 is virtually certain to be in that continuous stream too. Replaying it re-trips 1007 and locks the loop — which is exactly the bug this PR fixes, and the test on socket #2 demonstrates 0 audio messages + clean \`session.rotated\` confirms the loop is broken.
- Yes, this drops ~the reconnect-window of fresh speech (bounded by \`MAX_PENDING_AUDIO\` = 50 chunks ~= 1s). That's a deliberate trade vs. an unrecoverable 1007 loop. The PR description and the comment in code both call this out as the intended behavior.
- A finer-grained \"cutoff/generation\" approach as Codex suggests is a nice-to-have but not required to ship — the current behavior is strictly better than main (which already drops audio unconditionally on every reconnect).

**Gemini's Should-fix** (named constants for 1007/1011) is reasonable but pure style; it's a follow-up at most.

**Recommendation: ready to merge as-is.** Optional follow-up: tighten the comment on \`dropPendingAudioOnFlush\` to say \"queued audio captured during the reconnect window after a 1007\" rather than \"before that close,\" since that's what's actually being dropped.

PR #400, opened 2026-05-05
feat(desktop): disable image attachments on Grok Voice (NAN-718)

yagudaev:
## Why

Grok Voice has no visual modality. Screen-share is already gated for that reason (`ChatPage.tsx:603-640`) — the image-attach button wasn't. A user on Grok could attach a PNG/JPEG/WEBP that the adapter would silently drop. Mirroring the existing screen-share gating closes the gap.

Closes [NAN-718](https://linear.app/nano3labs/issue/NAN-718/disable-image-attachments-when-grok-voice-is-the-active-model).

## Approach

- `ChatComposer` gets a new optional `attachDisabledReason` prop. When set, the attach button stays visible but is disabled and the string becomes its tooltip. The button is wrapped in a `<span title=…>` so the tooltip still surfaces on a disabled control (browsers suppress `title` on disabled buttons natively).
- `ChatPage` computes the reason from `activeRealtimeModel` and passes it through. Tooltip copy mirrors the existing screen-share one shape-for-shape.
- **Defense in depth**: `attachments:pickImage` IPC handler also early-returns when `getActiveProvider() === 'xai'`, so a future UI bug can't bypass the renderer-side gate.
- **Mobile**: no image-attach UI exists today (verified by grepping for `Paperclip`, `ImagePicker`, `pickImage`, etc.), so nothing to gate there.

- [x] `yarn test` — 124 tests pass, 13 files (no behavior tests added — gating is plumbing-only)
- [ ] Manual: select Grok in Settings → return to chat → attach button is greyed, tooltip reads *"Image attachments are only available with Gemini Live. Grok Voice does not support image input."*
- [ ] Manual: switch back to Gemini → button enabled, original tooltip restored
- [ ] Manual: switch to Grok mid-call → button immediately gates without restart
- [ ] Manual: trigger the IPC directly (devtools `window.electronAPI.attachments.pickImage()`) on Grok → returns `{ ok: false, error: ... }` instead of opening the dialog

</details>

- The `attachDisabledReason` API matches the existing `screenShareTitle`/`screenShareDisabled` shape — keeps two adjacent feature-gates legible.
- Re-checking the active model in the IPC handler costs a single SQLite read (`getActiveProvider`); fine on a button-driven path.
- The renderer-side gate is the user-facing one; the IPC gate is just a safety net.

</details>

### DIFF
--- a/desktop/src/main/ipc-handlers.ts
+++ b/desktop/src/main/ipc-handlers.ts
@@ -361,6 +361,16 @@ export function registerIpcHandlers() {
   })
 
   ipcMain.handle('attachments:pickImage', async () => {
+    // Grok Voice has no visual modality. The renderer disables the
+    // attach button on this provider, but we re-check here so a future
+    // UI bug can't silently send an image the adapter won't relay.
+    if (getActiveProvider() === 'xai') {
+      return {
+        ok: false as const,
+        error:
+          'Image attachments are only available with Gemini Live. Grok Voice does not support image input.',
+      }
+    }
     const window = getMainWindow()
     const result = window
       ? await dialog.showOpenDialog(window, {
--- a/desktop/src/renderer/src/components/ChatComposer.tsx
+++ b/desktop/src/renderer/src/components/ChatComposer.tsx
@@ -17,11 +17,21 @@ import {
 export interface ChatComposerProps {
   onSubmit: (text: string) => void
   onAttach?: () => void
+  // When set, the attach button stays visible but is disabled and the
+  // string becomes its tooltip (used to gate image attachments on
+  // models that don't accept visual input — see ChatPage).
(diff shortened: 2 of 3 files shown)


yagudaev:
## Review responses

Codex reviewed `e8d932…` and flagged 1 Critical (Gemini timed out at 600s, no findings). Addressed in `…`:

| Finding | Reviewer | Severity | Status |
|---|---|---|---|
| Drag/drop and paste bypass the gate; `handleSendAttachments` then persists/forwards queued images | Codex | Critical | Fixed — `ingestFiles` and `handleSendAttachments` both check `attachDisabledReason` and surface the same inline error. The send-time check also covers the "queued on Gemini, switched to Grok before send" race |

No should-fix or nits raised.

PR #401, opened 2026-05-05
chore(main): release desktop 0.10.47

yagudaev:
:robot: I have created a release *beep* *boop*
---

### Features

* **desktop:** disable image attachments on Grok Voice (NAN-718) ([#400](https://github.com/yagudaev/voiceclaw/issues/400)) ([ddd436c](https://github.com/yagudaev/voiceclaw/commit/ddd436c32ec4e0840d5b4c6490f541e122f745d3))

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).

PR #403, opened 2026-05-05
fix: restore get-windows dep + escape MDX braces to unblock CI

yagudaev:
## Why

CI on `main` is broken. Every workflow that runs `yarn install --immutable` fails with `YN0028: The lockfile would have been modified by this install` — Desktop smoke and Deploy docs both die at the install step. The release-please bot's `0.10.47` commit (#401) stripped the `get-windows` dependency that #366 had added to `desktop/package.json`, but the regenerated `yarn.lock` still references it. CI refuses to drop the orphan entry, so install fails before any real work runs.

`screen-capture.ts` still imports `get-windows` (used to look up CGWindowID for window-source screen-share alignment), so the right fix is to put the dep back in `package.json` — not let the lockfile drop it.

Bonus: once install passes, the Deploy docs job moves on and dies on a separate pre-existing bug — `Connection to {provider} closed unexpectedly` in `troubleshooting.mdx`. MDX evaluates the `{...}` as a JS expression, throwing `provider is not defined` during prerender. Escaping the braces was a one-character change so I rolled it into the same PR rather than letting docs-deploy stay broken on main.

<details>
<summary>Test plan</summary>

- [x] `yarn install --immutable` passes locally on this branch
- [x] `cd docs && yarn build` completes (34 pages, no errors)
- [ ] CI on the PR: Desktop smoke green
- [ ] CI on the PR: Deploy docs green

</details>

<details>

- I considered the alternative of regenerating `yarn.lock` to match the (mistakenly) stripped `package.json`. Rejected: `desktop/src/main/screen-capture.ts` and `desktop/src/renderer/src/pages/ChatPage.tsx` both reference `get-windows`, so dropping the dep would compile-error the desktop build. The lockfile entry was correct; the `package.json` regression is what to fix.
- The release-please diff that removed it (`cfc74f1`) is just a version bump from `0.10.46` → `0.10.47` plus the dependency line vanishing. Looks like a release-please tooling artifact rather than an intentional drop. May be worth investigating release-please config separately so this doesn't happen again on the next release.
- Docs deploy has been red since #380 (Apr 30) — the MDX `{provider}` was introduced there. Just unlucky we didn't notice until now because docs-deploy was the second-class citizen behind the more obvious lockfile failure.

</details>

### DIFF
--- a/desktop/package.json
+++ b/desktop/package.json
@@ -22,6 +22,7 @@
     "archiver": "^7.0.1",
     "better-sqlite3": "^12.9.0",
     "electron-updater": "^6.6.2",
+    "get-windows": "^9.3.0",
     "posthog-js": "^1.372.1",
     "posthog-node": "^5.30.4",
     "react": "^19.1.0",
--- a/docs/src/content/docs/desktop/troubleshooting.mdx
+++ b/docs/src/content/docs/desktop/troubleshooting.mdx
@@ -111,7 +111,7 @@ When the realtime voice connection fails because of an upstream provider problem
 | Gemini API key invalid | Key was deleted or rotated | Click **Open Settings** → re-enter the key under Settings → Provider |
 | Gemini quota exceeded | Daily quota hit | Click **Open billing** → upgrade at aistudio.google.com/app/billing |
 | Gemini service is having issues | Google-side outage | Check status.cloud.google.com; try again later |
-| Connection to {provider} closed unexpectedly | Transient network drop | Start a new call |
+| Connection to \{provider\} closed unexpectedly | Transient network drop | Start a new call |
 
 The banner stays visible for the rest of the session so you can confirm the fix after returning from the billing page. Click **Dismiss** to hide it, or start a new call to reset it.
 


yagudaev:
## 🤖 Codex review

**Reviewer:** `codex review --base main` (codex-cli 0.128.0, model gpt-5.5)
**Scope:** PR #403 diff (desktop/package.json, docs/src/content/docs/desktop/troubleshooting.mdx)

### Findings

> The changes align the desktop manifest with an existing runtime import and lockfile entry, and the MDX escaping is appropriate for rendering a literal placeholder. I found no actionable correctness issues in the diff.

No correctness, security, or build-impact issues flagged. The two changes do exactly what the PR title says:

- `desktop/package.json`: re-adds `get-windows@^9.3.0` that release-please stripped in v0.10.47, restoring agreement with `screen-capture.ts` import and the lockfile entry.
- `docs/.../troubleshooting.mdx`: escapes literal `{provider}` braces so MDX renders the placeholder instead of trying to evaluate it as a JS expression during prerender.

**Verdict:** LGTM — safe to merge to unblock CI.

yagudaev:
## 🤖 Gemini review

**Reviewer:** `gemini -p` (gemini CLI, headless mode)
**Scope:** PR #403 diff (2 files, +2/-1)

The PR diff looks solid and addresses the issues as described. Here is the review summary:

### 1. Restoration of `get-windows`
- **Confirmation of Use:** I verified that `get-windows` is indeed imported and used in `desktop/src/main/screen-capture.ts` (line 2 and line 80). Without this dependency in `package.json`, the desktop app would fail to build or run as the module would be missing during bundling.
- **Version Consistency:** The added version `^9.3.0` matches the version currently present in `node_modules`. This ensures consistency with the local development environment and CI.
- **No Missed Callers:** A search across the `desktop/` directory confirmed that `src/main/screen-capture.ts` is the only active caller. There is also a mention in a comment within `src/renderer/src/pages/ChatPage.tsx`, but no direct import there (which is correct as it's a native module).

### 2. MDX Syntax Fix
- **Correct Escaping:** Escaping `{provider}` as `\{provider\}` in `docs/src/content/docs/desktop/troubleshooting.mdx` is the correct solution. In MDX, curly braces are interpreted as JavaScript expressions; escaping them ensures they are rendered as literal text, unblocking the documentation prerender.

### 3. Risk Assessment
- **Native Modules:** `get-windows` v9+ on macOS uses a standalone Swift binary rather than a Node-API addon. This means it doesn't strictly require `electron-rebuild` for macOS, making it lower risk than other native dependencies like `better-sqlite3`.
- **Lockfile:** While the diff doesn't show a `yarn.lock` update, adding a dependency that was previously present (and likely still in the lockfile as an orphaned entry) is low risk. If `yarn` is used, running `yarn install` will clean this up.

### Conclusion
The fix is surgical and correctly addresses both the CI blockage (missing dep) and the prerender failure (MDX syntax). No further issues were identified.

**Recommendation:** Approve and merge to unblock CI.

